PR #25, opened 2015-11-21
Handle empty "method" & "action" and submit via GET correctly

jwodder:
This fixes the library to correctly handle the following scenarios:

- The `"method"` attribute of a form may be left unspecified, in which case it defaults to GET.
- The `"action"` method of a form may be left unspecified, in which case it should be the URL of the source document, which in this case is (assumed to be) the `url` argument to `submit`.  (If both `"action"` and `url` are undefined, raise an error.)
- When submitting form data via GET, the data should be passed to `requests` in the `params` argument, not `data`.

Also fixed a couple minor coding things.

### DIFF
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -84,7 +84,7 @@ class Browser(object):
                 if i == 0 or "selected" in option.attrs:
                     data[name] = option.get("value", "")
 
-        if method == "get":
+        if method.lower() == "get":
             kwargs["params"] = data
         else:
             kwargs["data"] = data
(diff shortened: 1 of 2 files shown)


hickford:
Looks good to me, thanks.

mechanicalsoup/browser.py
```diff
@@ -5,7 +5,7 @@
 from .form import Form
 
 
-class Browser:
+class Browser(object):
```
hickford: Out of curiosity, what does this do?

jwodder: @hickford In Python 2, this is necessary to make `Browser` [a new-style class instead of an old-style class](http://stackoverflow.com/q/54867/744178).  (In Python 3, all classes are new-style, so explicitly inheriting from `object` is superfluous.)

hickford: Cool cheers

PR #31, opened 2016-01-10
added requests's custom adapter implementation

gsora:
This change will make us able to implement custom transport adapter, increasing possibilities of connection customization.

It's possible to pass to `Browser()` a dict containig the adaptee as key, and the adapter itself as value.
 
(see for reference: http://docs.python-requests.org/en/latest/user/advanced/#transport-adapters)


hickford:
Looks good to me, thanks

PR #34, opened 2016-02-21
Get rid of BeautifulSoup warning

mohan3d:
get rid of this warning,

UserWarning: No parser was explicitly specified, so I'm using the best available HTML parser for this system ("html.parser"). This usually isn't a problem, but if you run this code on another system, or in a different virtual environment, it may use a different parser and behave differently.


hickford:
Can we be sure every user will have html.parser ?

mohan3d:
isn't html.parser built-in python ?

PR #39, opened 2016-06-11
Fix incorrect PyPI version badge

orangain:
A version of PyPI badge is incorrect.

| Before | After |
|-----|----|
|  |  |


hickford:
Thanks

PR #41, opened 2016-08-22
browser: add a launch_browser method

moy:
For debugging, it's very convenient to be able to fire up a browser on
the current page.

### DIFF
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -3,6 +3,8 @@ import bs4
 from six.moves import urllib
 from six import string_types
 from .form import Form
+import webbrowser
+import tempfile
 
 
 class Browser(object):
@@ -115,3 +117,9 @@ class Browser(object):
         response = self.session.send(request)
         Browser.add_soup(response, self.soup_config)
         return response
+
+    def launch_browser(self, soup):
+        """Launch a browser on the page, for debugging purpose."""
+        with tempfile.NamedTemporaryFile(delete=False) as file:
+            file.write(soup.encode())
+        webbrowser.open('file://' + file.name)


moy:
Strange: I get a build failure for Python 3.2, but I tried a build for 73d7601 and it fails the same way (`cannot import name utils`), while the build was OK at the time it was pushed. I guess the latest version of requests does not support Python 3.2 anymore (which makes sense: people who want a recent version of Python use newer 3.x, and people who want an old version use 2.x).

hugovk:
PR https://github.com/hickford/MechanicalSoup/pull/44 fixes the Requests/3.2 failure by removing  support for that verison.

mechanicalsoup/browser.py
```diff
@@ -115,3 +117,9 @@ def submit(self, form, url=None, **kwargs):
         response = self.session.send(request)
         Browser.add_soup(response, self.soup_config)
         return response
+
+    def launch_browser(self, soup):
+        """Launch a browser on the page, for debugging purpose."""
+        with tempfile.NamedTemporaryFile(delete=False) as file:
+            file.write(soup.encode())
```
hickford: Is `soup.encode()` more useful than `response.body`?

moy: It is best to use the output of the HTML parser, because it launches the browser on the page as seen by the parser.

I had to fight bugs where Firefox was seeing something different from html.parser (probably because of a missing closing tag), and some input fields were seen outside a form by html.parser. Launching a browser on soup.encode() showed me directly the input field outside the form, and I could work around it. Launching the browser on the unparsed HTML wouldn't.

PR #50, opened 2017-01-17
example.py: restore changes lost by bad merge

### LINKED ISSUE #40: example.py seems broken
```
$ python example.py login password
  File "example.py", line 18, in <module>
    login_form = login_page.soup.select_one('#login form')
TypeError: 'NoneType' object is not callable
```

I tried adding `print(login_page.soup.select_one)` before the guilty line and it is indeed `None`.

moy:
ea5e675 (update example.py, 2015-06-30) introduced a bunch of
improvements to example.py, and most of these changes were accidentally
reverted by 4907e63 (Merge branch 'Xuefeng-Zhu-example', 2015-11-24).

In particular, the original commit introduced the creation of a
mechanicalsoup.Form() object based on a beautiful soup one, and this
change is needed for the code to work. Without this, one tries to call
method input() on a soup object instead of using the mechanicalsoup one.

Fixes #40.

### DIFF
--- a/example.py
+++ b/example.py
@@ -2,20 +2,21 @@
 import argparse
 import mechanicalsoup
 
-parser = argparse.ArgumentParser(description='Login to GitHub.')
+parser = argparse.ArgumentParser(description="Login to GitHub.")
 parser.add_argument("username")
 parser.add_argument("password")
 args = parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


hickford:
Cheers thanks

PR #53, opened 2017-01-31
[WIP/RFC] A bunch of new features from chamilotools

moy:
This PR is the result of moving code from https://gitlab.com/chamilotools/chamilotools to MechanicalSoup.

The two main features are:

* StatefulBrowser, which keeps track of the current URL, page and form for you.

* The ability to write `form["name"] = value` instead of form.{check,input,textarea}.

* Because the StatefulBrowser keeps track of the current form, we can even write `browser["name"] = value` instead of `form["name"] = value`.

See what it allows in the new [example.py](https://github.com/moy/MechanicalSoup/blob/stateful-browser/example.py). The main part is

```Python
browser.open("https://github.com")
browser.follow_link("login")
browser.select_form('#login form')
browser["login"] = args.username
browser["password"] = args.password
resp = browser.submit_selected()
```

It still needs:

* tests

* at least docstrings to document the API

I'll merge this soon if I don't get any comments/objection. Don't hesitate if you think anything's wrong or if you think something could be improved.


moy:
Oops, deleted the branch by mistake, but here is it again.

PR #54, opened 2017-01-31
Update .travis.yml with python 3.6

L1ghtn1ng:
Add python 3.6 to travis and remove old versions of python

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,13 +22,12 @@ setup(
         # Specify the Python versions you support here. In particular, ensure
         # that you indicate whether you support Python 2, Python 3 or both.
         'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 2.6',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.2',
-        'Programming Language :: Python :: 3.3',
         'Programming Language :: Python :: 3.4',
         'Programming Language :: Python :: 3.5',
+        'Programming Language :: Python :: 3.6'
     ],
 
     packages=['mechanicalsoup'],
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,9 @@
 language: python
 python:
-  - 2.6
   - 2.7
-  - 3.3
   - 3.4
   - 3.5
+  - 3.6
   - pypy
   - pypy3
 install:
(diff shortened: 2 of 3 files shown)


moy:
Is there a reason to drop support for old versions? In particular, Python 2.6 is not dead yet IMHO. Why not drop it if there's a good reason to, but then the good reason should be mentioned in the commit message.

L1ghtn1ng:
As far as I am aware no distribution is using these versions of python and for Mac users they should be using homebrew to get new versions of python as http://pyfound.blogspot.co.uk/2017/01/time-to-upgrade-your-python-tls-v12.html?utm_source=feedburner&utm_medium=twitter&utm_campaign=Feed:+PythonSoftwareFoundationNews+(Python+Software+Foundation+News)&m=1 this is going to be happening

________________________________

moy:
The fact that distro don't ship 2.6 anymore doesn't mean everybody has upgraded. I still use a production server with Python 2.4 for example. There are more RedHat/CentOS 6 out there than you seem to think.

I don't question the fact that most users should upgrade, but that doesn't seem to be a good reason for MechanicalSoup to drop support.

L1ghtn1ng:
Okay will fix when I get back from work

________________________________

moy:
Just to be clear: if there's a 2.7-only feature that you really want to use, I have no strong objection to dropping 2.6 support, but then it can be a preparatory patch for the commit introducing this Python construct. All I'm saying is: if we can keep support for 2.6 without a real overhead, let's keep it.

setup.py
```diff
@@ -22,13 +22,12 @@
         # Specify the Python versions you support here. In particular, ensure
         # that you indicate whether you support Python 2, Python 3 or both.
         'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 2.6',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.2',
```
moy: While you're at it, you can also remove 3.2 which isn't supported either. I'm not doing this to avoid a conflict for you.

Thanks,

moy:
OK, I committed a variant of this in 0798afffa03e27eb22032c9e0d5bbd37916cbe23 and applied my remark about 3.2 in setup.py in 778e019fa68c8187420db809cfa761fc7c54c016. We're all set, I'm closing the PR.

Thanks!

L1ghtn1ng:
Okay sorry about taking ages to fix this I had been ill and forgot and only remembered about this yesterday

________________________________

PR #57, opened 2017-02-06
New features (stateful browser, __setitem__, ...) from chamilotools

moy:
This PR is the result of moving code from https://gitlab.com/chamilotools/chamilotools to MechanicalSoup.

The two main features are:

* StatefulBrowser, which keeps track of the current URL, page and form for you.

* The ability to write `form["name"] = value` instead of form.{check,input,textarea}.

* Because the StatefulBrowser keeps track of the current form, we can even write `browser["name"] = value` instead of `form["name"] = value`.

See what it allows in the new [example.py](https://github.com/moy/MechanicalSoup/blob/stateful-browser/example.py). The main part is

```Python
browser.open("https://github.com")
browser.follow_link("login")
browser.select_form('#login form')
browser["login"] = args.username
browser["password"] = args.password
resp = browser.submit_selected()
```

It could get better testing and documentation, but I find the current state acceptable to be merged. It can be polished in-tree later.

I'll merge this soon if I don't get any comments/objection. Don't hesitate if you think anything's wrong or if you think something could be improved.

This is a resubmission of #53 which I closed by mistake and doesn't seem to be re-openable.

### DIFF
--- a/example.py
+++ b/example.py
@@ -1,0 +1,40 @@
+"""Example app to login to GitHub using the StatefulBrowser class."""
+
+from __future__ import print_function
+import argparse
+import mechanicalsoup
+from getpass import getpass
+
+parser = argparse.ArgumentParser(description="Login to GitHub.")
+parser.add_argument("username")
+args = parser.parse_args()
+
+args.password = getpass("Please enter your GitHub password: ")
+
+browser = mechanicalsoup.StatefulBrowser()
+# Uncomment for a more verbose output:
+# browser.set_verbose(2)
+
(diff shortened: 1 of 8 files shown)


hickford:
Nice. The original mechanize library was stateful http://wwwsearch.sourceforge.net/mechanize/

moy:
> The original mechanize library was stateful http://wwwsearch.sourceforge.net/mechanize/

That's where I come from ;-). And that's what I missed when I moved to MechanicalSoup seeing that mechanize was unmaintained and Python2-only.

PR #67, opened 2017-04-08
remove github formId from example.py/README.md

mohan3d:
running example.py raised  

```
LinkNotFoundError()    
mechanicalsoup.utils.LinkNotFoundError
```

used `"form"` instead of `"#login form"` since there is only one form in [github.com/login](https://github.com/login)

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ browser = mechanicalsoup.StatefulBrowser()
 
 browser.open("https://github.com")
 browser.follow_link("login")
-browser.select_form('#login form')
+browser.select_form('form')
 browser["login"] = args.username
 browser["password"] = args.password
 resp = browser.submit_selected()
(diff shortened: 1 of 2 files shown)


moy:
> running example.py raised
>
> LinkNotFoundError()    
> mechanicalsoup.utils.LinkNotFoundError

`example.py` works just fine here. I suspect an issue with your HTML parser (depending on which one is installed, BeautifulSoup will use a different one and it may yield different results). I tried with html.parser and then I can reproduce. I think the proper fix to your issue is what I committed in #68.

I'm personally in favor of keeping the big, scary warning of BeautifulSoup when the parser isn't specified explicitly, but I seem to disagree with @hickford here.

I don't like much the idea of removing this `#login`: it's good in the example to show a way to specify precisely which form you're referring to.

moy:
No answer, so I'm taking this as a "OK, #68 is the right way", but feel free to reopen or continue the discussion on PR #68 if this is not the case and you think of a better fix.

In any case, thanks for the report.

PR #70, opened 2017-05-02
Delete all but chosen submit; allow match on name.

dbrobins:


moy:
Looks good to me, but it would be really nice to have a testcase for this. Untested features tend to be broken silently rather easily ...

I'm merging this as-is, but it would be cool if you could provide a test in another PR.

Thanks!

dbrobins:
Sure, I'll add a test later this week.

PR #71, opened 2017-05-05
Test (fixed) form choose_submit method.

dbrobins:


moy:
Hi,

As you may have noticed, the Travis job fails with 'No module named requests_mock'. You need to adapt .travis-ci.yml to install requests_mock before running the tests.

Thanks,

moy:
OK, there's one Travis-CI failure, but it's not because of your changes: the master branch fails the same test too (it didn't by the time it was commited, but I restarted the test and it [fails](https://travis-ci.org/hickford/MechanicalSoup/jobs/228051876) now, probably because of [Travis-CI's latest update](https://blog.travis-ci.com/2017-05-04-precise-image-updates)). I'm merging this, and we should take care of the Travis failure too, obviously.

PR #72, opened 2017-06-07
*browser.py: allow raising LinkNotFound on 404 error

moy:
Ported from chamilotools (new tests added).


moy:
Should be ready to merge now (Travis fails on pypy3 because of an issue on Travis's side, but not on ours)

PR #79, opened 2017-08-17
Instructions on how to setup virtualenv for both Python2 & Python3

eoghanmurray:
I found the https://github.com/hickford/MechanicalSoup#development section a bit sparse.
If you've got a 'dev' version of mechanicalsoup and a system installed one that is maybe being used by other projects, it's not clear how to test the dev version separately, or how to test against multiple versions of Python.  Also, running `py.test` inside a .virtualenv didn't work for me as that looks at the system wide version I think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Development
 
 [![Build Status](https://travis-ci.org/hickford/MechanicalSoup.svg?branch=master)](https://travis-ci.org/hickford/MechanicalSoup)
 
-You can develop against multiple versions of Python using virtualenv:
+You can develop against multiple versions of Python using [virtualenv](https://packaging.python.org/tutorials/installing-packages/#creating-virtual-environments):
 
     python3 -m venv .virtual-py3 && source .virtual-py3/bin/activate
     pip install pytest flake8 requests_mock
--- a/README.md
+++ b/README.md
@@ -91,9 +91,27 @@ Development
 
 [![Build Status](https://travis-ci.org/hickford/MechanicalSoup.svg?branch=master)](https://travis-ci.org/hickford/MechanicalSoup)
 
-### Tests
+You can develop against multiple versions of Python using virtualenv:
 
-    py.test
+    python3 -m venv .virtual-py3 && source .virtual-py3/bin/activate
+    pip install pytest flake8 requests_mock
+and
+
+    virtualenv -p python2 --no-site-packages .virtual-py2 && source .virtual-py2/bin/activate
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -91,9 +91,27 @@ Development
 
 [![Build Status](https://travis-ci.org/hickford/MechanicalSoup.svg?branch=master)](https://travis-ci.org/hickford/MechanicalSoup)
 
-### Tests
+You can develop against multiple versions of Python using virtualenv:
+
+    python3 -m venv .virtual-py3 && source .virtual-py3/bin/activate
```
moy: I have no experience with virtualenv, but this doesn't work for me:

```
$ python3 -m venv .virtual-py
Error: Command '['/home/moy/dev/MechanicalSoup/.virtual-py/bin/python3', '-Im', 'ensurepip', '--upgrade', '--default-pip']' returned non-zero exit status 1
```

OTOH, the command I find [here](http://docs.python-guide.org/en/latest/dev/virtualenvs/) does work:

```
$ virtualenv -p /usr/bin/python3 my_project
...
Installing setuptools, pip, wheel...done.
$ source ./my_project/bin/activate
```

Is there any reason to use `python3 -m venv` instead of `virtualenv`? My setup may well be broken, but as-is I can't confirm that your instructions are correct.

eoghanmurray: Hmmm do you have a version of Python3 less than 3.3?
https://packaging.python.org/tutorials/installing-packages/#creating-virtual-environments

I think the only advantage of `python3 -m venv` is that you can skip the `pip3 install virtualenv` step.

moy: It was on a Debian oldstable (I don't have it at hand, but should be 3.4.2 according to Debian's website). I tested again on another machine and your version works.

However, you're suggesting `python3 -m venv` for Python3 and then `virtualenv -p python2` for Python2. I think it'd be better to use the same syntax (no preference on which one on my side) for both.

Also, the link you just gave is a good one, it would be nice to include it in your explanation (for example as a hyperlink on the first occurence of "virtualenv").

Thanks,

eoghanmurray:
Thanks for the merge - I think `python3 -m venv` is the recommended way for Python3 as virtualenv has been accepted as an integratal part of Python.  Hopefully people will be able to figure out they can do `virtualenv -p python3 --no-site-packages .virtual-py3` if the new way doesn't work, but I'm happy to leave the `python3 -m venv` way there as I hadn't heard of it til I did some research!

PR #85, opened 2017-09-12
browser.py: add a direct interface to the cookiejar

hemberger:
* The `Browser.{set,get}_cookie` methods provide a way to set and
  get the requests cookiejar without needing to know the details
  of the requests module.

* Improve consistency with mechanize, which also provides a
  `Browser.set_cookie` method.


moy:
The commit message talks about `{set,get}_cookie` but the code defines `{set,get}_cookiejar`. Can you fix the commit message?

hemberger:
Good catch, thanks!

PR #87, opened 2017-09-12
Add unit test of `StatefulBrowser.links` method

hemberger:


moy:
Thanks. It would be even better IMHO to test on a page containing several links (to check that we're able to return a non-singleton). But that's definitely an improvement anyway.

Issue #92, opened 2017-09-23
What does `StatefulBrowser.submit_selected`'s `btnName` argument do?


hemberger:
There is no documentation of the `btnName` argument of the `StatefulBrowser.submit_selected` method, and it simply gets forwarded to the `requests.Request` constructor in a data dict (the [documentation](http://docs.python-requests.org/en/master/api/#requests.Request) for this argument in the `requests` module is not very illuminating either).

I have an example form that looks like this:
```html
<form class="standard" method="POST" action="?sn=2e1b94b5">
    <input class="submit" name="action" type="submit" value="View"/>
    <input class="submit" name="action" type="submit" value="Save"/>
    <input type="checkbox" value="on" name="id[788441]"/>
    <input type="checkbox" value="on" name="id[788558]"/>
</form>
```

Thinking `btnName` meant "button name" (as in, the value of the submit element to be clicked), I tried the following code:
```python
def SelectCheckboxes(br, names, submit_value):
  br.select_form('form')
  br.get_current_form().check({name: "on" for name in names})
  br.submit_selected(btnName=submit_value)

SelectCheckboxes(br, ['id[788441]'], "View")
```

The above code submits using the "Save" element instead of "View". Unsurprisingly, the code works correctly with the following change to `SelectCheckboxes`:
```diff
-  br.submit_selected(btnName=submit_value)
+  br.get_current_form().choose_submit(submit_value)
+  br.submit_selected()
```
But the existence of the `choose_submit` method further disguises what `btnName` is actually supposed to be. Any clarification would be greatly appreciated!

hemberger:
Upon further investigation, there are three problems here:
1. There is a bug in the return value of `Form.choose_submit` where it was returning `True` when it hadn't found the specified element.

2. `Form.choose_submit` only has the ability to select input element _names_, not input element _values_. Because the return value was incorrect (problem 1), I didn't notice this was happening. Note: `mechanize` can select input elements by name using the `label` argument (i.e. `br.submit(label=submit_value)`), so perhaps it can be added as a feature here.

[...]

PR #94, opened 2017-09-25
Fix methods that choose between multiple submit elements

hemberger:
`Form.choose_submit`:
* Fix the logic that decides when a submit element matches the input argument. The matching was done correctly, but the bookkeeping about the match was not.
* Improve safety by replacing the boolean return value with a raised `LinkNotFoundError` exception unless exactly one match is found.

`StatefulBrowser.submit_selected`:
* Fix the `btnName` argument so that the POST data is correctly generated in the case where the form has multiple submit elements.

See commit messages and issue #92 for more details.

Minor changes:
* Add or improve method docstrings.
* Remove manual (non-`pytest`) infrastructure for running tests due to the addition of `pytest` features.
* Additional test cases and code de-duplication of `test_choose_submit`.

### DIFF
--- a/tests/test_browser.py
+++ b/tests/test_browser.py
@@ -1,8 +1,9 @@
 import mechanicalsoup
+import sys
 from bs4 import BeautifulSoup
 import tempfile
 from requests.cookies import RequestsCookieJar
-
+import pytest
 
 def test_submit_online():
     """Complete and submit the pizza form at http://httpbin.org/forms/post """
@@ -135,3 +136,6 @@ def test_get_cookiejar():
     jar = browser.get_cookiejar()
     assert jar.get('k1') == 'v1'
     assert jar.get('k2') == 'v2'
+
+if __name__ == '__main__':
+    pytest.main(sys.argv)
--- a/tests/test_browser.py
+++ b/tests/test_browser.py
@@ -135,12 +135,3 @@ def test_get_cookiejar():
     jar = browser.get_cookiejar()
     assert jar.get('k1') == 'v1'
     assert jar.get('k2') == 'v2'
-
-if __name__ == '__main__':
-    test_submit_online()
-    test_build_request()
-    test_prepare_request_file()
-    test_no_404()
-    test_404()
-    test_set_cookiejar()
-    test_get_cookiejar()
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -1,4 +1,5 @@
 import mechanicalsoup
+import sys
 import requests_mock
 import pytest
 try:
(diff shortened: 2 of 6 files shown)


tests/test_form.py
```diff
@@ -98,35 +99,27 @@ def text_callback(request, context):
         mock.register_uri('POST', url + '/post', text=text_callback)
     return mechanicalsoup.StatefulBrowser(requests_adapters={'mock': mock}), url
 
-def test_choose_submit():
-    browser, url = setup_mock_browser(
-        expected_post=[
+@pytest.mark.parametrize("expected_post", [
+    pytest.param(
```
moy: Travis-CI says "AttributeError: 'module' object has no attribute 'param'" for Python 2.6, 3.3 and 3.4.

Any way to get this to work for all Python versions? Not sure whether it would work, but perhaps just adding a "pip install --update pytest" in .travis.yml could do the trick?

tests/test_browser.py
```diff
@@ -135,12 +135,3 @@ def test_get_cookiejar():
     jar = browser.get_cookiejar()
     assert jar.get('k1') == 'v1'
     assert jar.get('k2') == 'v2'
-
-if __name__ == '__main__':
```
moy: I'd like to keep a way to call the script without using the executable `pytest` (just for simplicity, people see a `*.py` file and may want to run `python ./test_file.py` or just `./test_file.py`).

AFAICT, calling `pytest.main(__file__)` would do the trick.

hemberger: That sounds like a fair compromise. I'll give it a try!

moy: Actually, `pytest.main(sys.argv)` is better (first, recent pytest warn when you pass a string to main and my initial version didn't allow `./test_foo.py -v` for example).

Search for better option in progress here: https://stackoverflow.com/questions/35353771/invoke-pytest-from-python-for-current-module-only

tests/test_form.py
hemberger: Thanks for the suggestion! Using `pip install --upgrade pytest` provides a sufficiently recent version of pytest.

tests/test_browser.py
hemberger: I ran into this suggestion as well, and it appears to provide all the necessary functionality. Thanks!

moy:
Looks good to me after a quick look.

Can you `git rebase -i` to squash your 3 last fixups into the relevant commits? If not I'll do that for you when merging/reviewing.

Thanks,

moy:
OK, I'll do the rebase for you actually (I'll use your `import pytest` for something else).

moy:
Merged, thanks for your contribution.

PR #97, opened 2017-09-28
Drop unsupported Python 2.6 and 3.3

hugovk:
Python 2.6 has been at end-of-life as of 2013-10-29, and 3.3 has one day left, until 2017-09-29.

https://snarky.ca/stop-using-python-2-6/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: python
 python:
-  - 2.6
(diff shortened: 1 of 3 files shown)


moy:
I'll make one last release supporting these old Python version and merge this after.

Python 2.6 is the version available on RHEL 6 for example. There are still quite a few users out there.

PR #110, opened 2017-10-12
.codecov.yml: disable PR comments

hemberger:
The Codecov PR comments are too verbose and do not add anything that
the Codecov GitHub App does not provide in the integrated checks.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,2 +1,2 @@
-comment:
-  require_changes: true
+# Don't post a comment on pull requests
+comment: off


moy:
The `require_changes` was an attempt to make the bot a bit quieter, but I must have done something wrong since we still got the report for #108 which didn't change coverage. I thought maybe comments could help non-regular contributors (who get feedback even if they don't know where to click to get the codecov report), but I also agree it's a bit too noisy (getting 2 emails per PR is a bit boring).

So, why not disable it completely, yes.

Issue #109, opened 2017-10-12
Development workflow and test dependencies


hemberger:
Somewhat related to Issue #100, the development dependencies listed in `README.md` can easily get out of sync with those used by the tests (and actively tested against by Travis-CI), e.g. see PR #108. Perhaps we should move to the "requirements file" scheme with a `requirements.txt` and `tests/requirements.txt` (or `test_requirements.txt`, which for better or worse is consistent with what Tox uses).

The thing I normally dislike about requirements files is that it makes the manual installation more complicated (you need to include `-r requirements.txt`). However, it seems simple to programmatically read these files in `setup.py`: https://stackoverflow.com/a/29871401.

Assuming this works as advertised, does this seem like a reasonable resolution to the desync problem?

moy:
I was thinking the same. Actually, it's a pity we can't just run something like `python setup.py install-dependencies` or so, but I didn't find a better way than parsing a requirement file as you suggest.

IOW, I'm all for the proposed change.

moy:
PS: I think it would make sense to have both test-dependencies and non-test dependencies in two requirement files. For example, readthedocs needs to install bs4 to generate the doc properly and it accepts a requirement file. For now I've ticked the "Install your project inside a virtualenv using setup.py install" box in the advanced settings, and the installation installs the dependencies as a side-effect, but it would be cleaner to provide the requirement file.

hemberger:
Does anything more need to be done with the requirements files before closing this issue?

moy:
I think it can be closed.

PR #117, opened 2017-10-13
Switch to requirements.txt files

hemberger:
Addresses issue #109.

Move install and test dependencies out of `setup.py` and into
`requirements.txt` and `tests/requirements.txt` respectively.

To avoid duplicating the list of dependencies, we read these
files in `setup.py` to build the `install_requires` and
`tests_require` lists.

In the development workflow instructions, we can recommend
installing all dependencies with pip using these files to avoid
yet another place where the list of dependencies would otherwise
be duplicated.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,13 @@ from setuptools import setup  # Always prefer setuptools over distutils
 from codecs import open  # To use a consistent encoding
 from os import path
 
+
+def requirements_from_file(filename):
+    """Parses a pip requirements file into a list."""
+    return [line.strip() for line in open(filename, 'r')
+            if line.strip() and not line.strip().startswith('--')]
+
+
 here = path.abspath(path.dirname(__file__))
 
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -2,6 +2,13 @@
 from codecs import open  # To use a consistent encoding
 from os import path
 
+
+def requirements_from_file(filename):
+    """Parses a pip requirements file into a list."""
+    return [line.strip() for line in open(filename, 'r')
+            if line.strip() and not line.strip().startswith('--')]
```
moy: For completeness, you may want to add comments management (strip `#.*`), but we don't really need it for now.

hemberger: Agreed. For future reference, if we want to parse a fully generic requirements file, the syntax is outlined [here](https://pip.pypa.io/en/latest/reference/pip_install/#requirements-file-format).

PR #120, opened 2017-10-17
Convert README to RST and use it on PyPi

moy:


hemberger:
As a first step, it looks good to me! I haven't verified that it renders as expected, but I assume you have ;)

PR #122, opened 2017-10-17
Restructure docs, more examples

moy:


moy:
GitHub rendering available here: https://github.com/moy/MechanicalSoup/tree/restructure-docs

Rendering of the project description (processed README.rst): http://matthieu-moy.fr/tmp/2017/README.html

hemberger:
I think it would be nice to keep the badges in the README (whether in the Development section as before, or all of them at the top, I'm not sure). It's what a lot of people look at to quickly assess a project, so hidden under a link is to our disadvantage.

hemberger:
I like how the two installation methods have been put in the same section, however it is now unclear that the python versions badge only refers to the version in PyPI. I'm assuming it's not possible to automatically build and package a `mechanicalsoup-dev` module for PyPI (or something else that would give us a nice description of the versions supported in both the "stable" and "latest" builds)?

moy:
Badges re-added.

About the version being installed, what do you think of 125e03b ?

Pypi accepts [pre-releases](https://packaging.python.org/tutorials/distributing-packages/#pre-release-versioning), but I don't know any way to deploy some automatically, and you can't re-upload an already uploaded version so we'd need an automatic version number increment.

OTOH, you can already install from GitHub. I'm documenting this with 4da3f3c (instructions do not work with master, but they do work with moy/MechanicalSoup so they should work once this PR is merged).

hemberger:
I think the version info you have now is sufficient. We can tweak it later if we feel that it's misleading.

PR #123, opened 2017-10-18
form.py: refactor set method

hemberger:
Instead of having Form.set implement textarea/check/select/input
itself, call the methods that already implement these. This change
reduces code duplication and prevents the two implementations
from getting out of sync (which had already happened with Form.check).

A new exception InvalidFormMethod is derived from LinkNotFoundError
and used in the methods called by Form.set to catch when the method
doesn't apply to the particular element being set. This is how
Form.set does its element type deduction now.

All methods called from Form.set are now prefixed with "set_".

Form.set
  * Is now a wrapper around Form.set_*.
  * Retains its original behavior.

Form.check
  * Is now a wrapper around Form.set_{checkbox,radio}.
  * Retains its original behavior for backwards compatibility, even
    though its handling of checkboxes is different than Form.set.

Form.set_radio
  * Factored out of Form.check.
  * Split into its own method because the design for radios is
    sufficiently different (and less complicated) than checkboxes
    that it warrants its own function for clarity.

Form.set_checkbox
  * Factored out of Form.check to implement new behavior that is
    consistent with Form.set.
  * Provides a "uncheck_others" argument to improve clarity about
    how it handles unspecified boxes. Form.set uses the default
    (True) and Form.check uses False (for backwards compatibility).
    The behavior in Form.set only changed recently (fc98922b94f67).
  * Implements the True/False values previously in Form.set but
    not in Form.check.

Form.set_select
  * Factored out of Form.set

Form.set_input
  * Renamed from Form.input (which is now an alias).

Form.set_textarea
  * Renamed from Form.textarea (which is now an alias).

The aliases and Form.check should be deprecated.

____________________________________________
TODO items:
- [x] better name for InvalidFormMethod exception
- [x] add tests of untested set_radio branches
- [x] turn off autodocs for method aliases
- [x] better name for "uncheck_others" argument to set_checkboxes
- [x] InvalidFormMethod is not showing up in Sphinx
- [x] Document changes in the changelog

### DIFF
--- a/docs/ChangeLog.rst
+++ b/docs/ChangeLog.rst
@@ -26,6 +26,16 @@ Main changes:
 
     pip install -r requirements.txt -r tests/requirements.txt
 
+* The ``Form`` class was restructured and has a new API. The behavior of
+  existing code is unchanged, but a new collection of methods has been
+  added for clarity and consistency with the ``set`` method:
+
+  - ``set_input`` deprecates ``input``
+  - ``set_textarea`` deprecates ``textarea``
+  - ``set_select`` is new
+  - ``set_checkbox`` and ``set_radio`` together deprecate ``check``
+    (checkboxes are handled differently by default)
+
 Bug fixes
 ---------
 
--- a/mechanicalsoup/__init__.py
+++ b/mechanicalsoup/__init__.py
@@ -1,8 +1,8 @@
 from .utils import LinkNotFoundError
 from .browser import Browser
-from .form import Form
+from .form import Form, InvalidFormMethod
 from .stateful_browser import StatefulBrowser
 from .__version__ import __version__
 
 __all__ = ['StatefulBrowser', 'LinkNotFoundError', 'Browser', 'Form',
-           '__version__']
+           'InvalidFormMethod', '__version__']
--- a/mechanicalsoup/form.py
+++ b/mechanicalsoup/form.py
@@ -2,113 +2,226 @@ from .utils import LinkNotFoundError
 from bs4 import BeautifulSoup
 
 
+class InvalidFormMethod(LinkNotFoundError):
+    """This exception is raised when a method of :class:`Form` is used
+    for an HTML element that is of the wrong type (or is malformed).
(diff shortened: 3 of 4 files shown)


hemberger:
Regarding the duplicate documentation for the method aliases, I think we have three options:
1. Suppress the docs by putting the methods in another file and telling Sphinx to exclude that file
2. Suppress the docs by making the aliases part of a base class that `Form` inherits from
3. Add explicit documentation for each alias stating that it's deprecated. This would change something like
```
textarea = set_textarea
```
into
```
def textarea(self, *args **kwargs):
    """DEPRECATED. Use :func:`~Form.set_textarea` instead."""
    self.set_textarea(*args, **kwargs)
```
(Possibly even with a `print("Form.textarea is deprecated. Please use Form.set_textarea instead.")` as well if we want to be super explicit about the deprecation.)

Thoughts?

moy:
Untested, but I think you can also:

4) do the `self.alias = self.actual_function` in the constructor.
5) use Sphinx autodoc's `:exclude-members:` and point it to aliases.

I think 4) would probably be the simplest.

About a `print` warning, I'd rather avoid at least for now: it's not a good idea to encourage people to change their code now, as it means their new code won't work with old MechanicalSoup. It's less serious for a small lib like MechanicalSoup that people are unlikely to have installed by the system, but in general a good deprecation flow is: add the new name and keep a silent alias. A long time after, start warning. A very long time after, remove the old alias (or not).

hemberger:
For every list I write that I thought was exhaustive, you have tricks up your sleeve to add to them! :)

4) works, and I agree it seems the simplest. Thanks!

hemberger:
I am happy with this PR now, but would like a 2nd pair of eyes before merging since it is a fairly significant change.

moy:
I prefer a history nicely split into logical commits (for example, you're introducing aliases in one commit and moving them around later in history, you could have a commit that introduces them at the right place directly, `git rebase -i` is your friend), but it'd be too much work for little benefit to do that now. Let's merge this.

PR #125, opened 2017-10-18
Faq

moy:


docs/faq.rst
```diff
@@ -59,3 +59,60 @@ This error can occure within requests' ``session.py`` when called by
 the destructor (``__del__``) of browser. The solution is to
 call :func:`~mechanicalsoup.Browser.close` before the end of life of
 the object.
+
+How do I get debug information/logs
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To unerstand what's going on while running a script, you have two
```
hemberger: Typo in "understand" here.

docs/faq.rst
```diff
@@ -59,3 +59,60 @@ This error can occure within requests' ``session.py`` when called by
...
+How does MechanicalSoup compare to the alternatives?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+There are other libraries with the same purpose as MechanicalSoup:
+
+* `mechanize <http://wwwsearch.sourceforge.net/mechanize/>`__, an
+  ancestor of MechanicalSoup (getting its name from the Perl mechanize
+  module). It was a great tool, but doesn't support Python 3, isn't
+  maintained anymore, and its documentation is poor. You shouldn't use
+  it anymore.
```
hemberger: As of this year, there is a new maintainer of Mechanize. There are extensive docs now (https://mechanize.readthedocs.io/en/latest/), but he has no intention of porting to python3 (https://github.com/python-mechanize/mechanize/issues/9). So I would agree that it shouldn't be used anymore, but the other statements might be a bit harsh.

hemberger:
The "logging" FAQ is really helpful. I didn't know about that functionality previously!

docs/faq.rst
moy: Oups, and it was not even the only typo. A spell checker does help ;-).

docs/faq.rst
moy: Indeed, thanks for the pointers.

eli-schwartz: Not really true, this was always more of the "patches welcome" variety. And mechanize has finally been updated to support python3 on master.

PR #131, opened 2017-10-19
Improve docstrings

hemberger:
In this PR:
* Add missing docstrings
* Improve clarity of existing text
* Add many sphinx function/class links
* Add some sphinx `:param:` or `:return:` docs
* More links to external resources when we use another object's API

Things not in this PR:
* Complete coverage of `:param:` and `:return:` (these can be a bit clunky for simple functions, but we may decide that we want to include them in the end)
* Example code

I think there's still a lot of room for improvement, but it's a good first pass. This also cleans up the internal interface a bit where I noticed areas for improvement while documenting.

- [x] StatefulBrowser
- [x] Browser
- [x] Form

### DIFF
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -77,6 +77,10 @@ class Browser(object):
         """Straightforward wrapper around `requests.Session.request
         <http://docs.python-requests.org/en/master/api/#requests.Session.request>`__.
 
+        :return: `requests.Response
+            <http://docs.python-requests.org/en/master/api/#requests.Response>`__
+            object with a *soup*-attribute added by :func:`add_soup`.
+
         This is a low-level function that should not be called for
         basic usage (use :func:`get` or :func:`post` instead). Use it if you
         need an HTTP verb that MechanicalSoup doesn't manage (e.g. MKCOL) for
@@ -89,6 +93,10 @@ class Browser(object):
     def get(self, *args, **kwargs):
         """Straightforward wrapper around `requests.Session.get
         <http://docs.python-requests.org/en/master/api/#requests.Session.get>`__.
+
+        :return: `requests.Response
+            <http://docs.python-requests.org/en/master/api/#requests.Response>`__
+            object with a *soup*-attribute added by :func:`add_soup`.
         """
         response = self.session.get(*args, **kwargs)
         if self.__raise_on_404 and response.status_code == 404:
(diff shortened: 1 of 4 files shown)


mechanicalsoup/form.py
```diff
@@ -256,10 +256,10 @@ def new_control(self, type, name, value, **kwargs):
         self.form.append(control)
         return control
 
-    def choose_submit(self, el):
```
moy: Note that you're potentially breaking backward compatibility for people using `choose_submit(el=foo)`. `choose_submit` was introduced relatively recently so I don't think it has so many users and we can hope that nobody wrote this `el=foo`, so I'm OK with this.

mechanicalsoup/stateful_browser.py
```diff
@@ -251,7 +251,7 @@ def follow_link(self, link=None, *args, **kwargs):
         Before raising, if debug is activated, list available links in the
         page and launch a browser.
 
-        :return: Forwarded from :func:`open`.
+        :return: Forwarded from :func:`open_relative`.
```
moy: Since open_relative actually forwards from open, the old doc was also correct. Probably we can keep the old version here to avoid a double-indirection for the user.

mechanicalsoup/browser.py
```diff
@@ -56,26 +74,32 @@ def set_user_agent(self, user_agent):
...
+        need an HTTP verb that MechanicalSoup doesn't manage (e.g. MKCOL) for
         example.
         """
         response = self.session.request(*args, **kwargs)
         Browser.add_soup(response, self.soup_config)
         return response
 
     def get(self, *args, **kwargs):
+        """Straightforward wrapper around `requests.Session.get
+        <http://docs.python-requests.org/en/master/api/#requests.Session.get>`__.
```
moy: Probably this should mention that in addition to requests's get, we remember the soup object in the browser.

hemberger: Agreed! I've added more complete return type information to the Browser methods that return a Response+Soup.

mechanicalsoup/form.py
hemberger: Why is it so painful to do the right thing! ;)

I agree that most people would have written `choose_submit(foo)` instead of `choose_submit(el=foo)`, and `el` is a painfully out-of-place variable name, but your instinct to preserve backwards compatibility is a good one. For now, I'll separate this change out into another PR so we can think on it for a bit longer.

mechanicalsoup/stateful_browser.py
hemberger: Since we have a lot of methods that are wrappers for each other, there's more double-indirection going on than just this instance. I'm going to keep it how it is for now for the sake of consistency, but as we iterate on this documentation again, I would be surprised if it didn't get modified.

mechanicalsoup/form.py
moy: At worse, you can keep `el=None` as argument, and do stg like

    if el is not None:
        print("el is deprecated, please use submit=... instead")
        if submit is not None:
            raise ...
        submit = el

We'd really need that if we were a 10,000,000 users lib. Probably overkill here. Actually, if someone wrote `el=...` somewhere in a script, then this script doesn't work but it raises a reasonable error and should be straightforward to fix.

Backward compatibility is painful for us, but it's really annoying to run a 2-year old script and notice it doesn't work anymore.

PR #132, opened 2017-10-20
form.py: rename argument of choose_submit

hemberger:
The name `el` was not sufficiently descriptive or clear. Instead,
we name it `submit` because it is the Tag of the submit element
or its name-attribute.

---------------------------
We are deciding whether or not to make this change, because it could in theory break backwards compatibility if someone were to have written `choose_submit(el=foo)`. See previous discussion in PR #131.

### DIFF
--- a/mechanicalsoup/form.py
+++ b/mechanicalsoup/form.py
@@ -257,10 +257,10 @@ class Form(object):
         self.form.append(control)
         return control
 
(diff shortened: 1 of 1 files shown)


moy:
OK, this PR keeps bugging me (having 0 pending PR is so nice ...). Let's merge this.

PR #137, opened 2017-10-20
Release checklist

moy:
I'd like to release a 0.9, if only to get the PyPi page updated (there are not many user-facing changes in the code, even though there's been a real overhaul inside).

In the process of avoiding the single point of failure, I think it would be nice to let @hemberger do it (enjoy your new priviledges ;-) ).

To help with this, here's a list of what I usually do (well, I didn't do the "Update the version number to a ``x.y.dev`` number in ``mechanicalsoup/__version__.py``" step last time, but probably I should have). Not sure CONTRIBUTING.rst is the best place, but I picked this to avoid creating yet-another-file that no one would read.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -65,6 +65,45 @@ Installation should be as simple as:
 
     python setup.py install
 
+Release Checklist
+-----------------
+
+Releases can be done only by people with sufficient priviledges on
+GitHub and PyPI. Things to do are:
+
+At each release:
+
+- Make sure all changes are documented in ``docs/ChangeLog.rst``
+- Update the version number to $v in
+  ``mechanicalsoup/__version__.py``
+- Remove the ``(in development)`` mention in ``docs/ChangeLog.rst``.
+- git commit -m "Release $v"
+- git tag v$ver
+- git push origin master v$v
+- Visit the `release page on GitHub
+  <https://github.com/MechanicalSoup/MechanicalSoup/releases>`__, copy
+  the relevant section from ``docs/ChangeLog.rst`` to the release
+  page.
+- ``python setup.py sdist upload -r pypi`` or ``twine upload``. This
+  requires a ``~/.pypirc`` with::
+
+    [distutils]
+    index-servers =
+        pypi
+    
+    [pypi]
+    username:<user>
+    password:<password>
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -65,6 +65,45 @@ Installation should be as simple as:
 
     python setup.py install
 
+Release Checklist
+-----------------
+
+Releases can be done only by people with sufficient priviledges on
+GitHub and PyPi. Things to do are:
```
hemberger: Completely trivial, but "PyPi" should be "PyPI". Feel free to ignore. :)

tests/test_browser.py
```diff
@@ -77,6 +78,7 @@ def test_build_request():
     request = browser._prepare_request(form)
     assert "application/x-www-form-urlencoded" in request.headers[
         "Content-Type"]
+    browser.close()
```
hemberger: I'm confused about all these calls to `close()`. Isn't this called when `browser` is destructed (due to going out of function scope)?

hemberger:
By the way, I contacted @hickford about a week ago asking if he could include me as a contributor in PyPI, but he hasn't gotten back to me yet. We will need to ping him again if we want me to perform the next release.

tests/test_browser.py
moy: I don't fully understand the issue, but the issue does occur: http://mechanicalsoup.readthedocs.io/en/latest/faq.html#referenceerror-weakly-referenced-object-no-longer-exists

The warning is essentially harmless, I saw a lot of them with tests passing (i.e. `python setup.py test` showed the warning and still concluded that the tests were passing) before I added the `.close` calls. But some of the tests check stderr, and these tests were flaky. See e.g. https://travis-ci.org/MechanicalSoup/MechanicalSoup/builds/290550991 for a failure.

More info here:
https://github.com/requests/requests/issues/2303
Essentially the conclusion is "don't do anything complex in `__del__`" and it seems `session.close()` is already too complex. But I may well have missed something.

moy:
OK, let me know if you want me to do the push to PyPI or if you prefer waiting a bit.

hemberger:
I'd prefer to wait, if that's okay with you. How about we merge this PR and then set up a 0.9 release checklist in Issues?

PR #138, opened 2017-10-20
Prefer https over http in documentation links

hemberger:
Does it matter if we still use http when connecting to http://httpbin.org for tests? (should we be testing both http and https?)


moy:
We could test http://httpbin.org in some tests and https:// in others, to make sure both work. But that's essentially the job of resquests: if we find an https-related breakage, it's 99% sure a request-issue.

PR #139, opened 2017-10-20
browser: implement __enter__ and __exit__

moy:
We implement the same methods as request.session, allowing code like

  def test_with():
      with mechanicalsoup.StatefulBrowser() as browser:
          browser.open(url)
	  # ...
      # implicit call to browser.close() here.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -60,6 +60,15 @@ the destructor (``__del__``) of browser. The solution is to
 call :func:`~mechanicalsoup.Browser.close` before the end of life of
 the object.
 
+Alternatively, you may also use the ``with`` statement which closes
+the browser for you::
+
+  def test_with():
+      with mechanicalsoup.StatefulBrowser() as browser:
+          browser.open(url)
+          # ...
+      # implicit call to browser.close() here.
+
 How do I get debug information/logs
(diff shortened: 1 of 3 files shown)


docs/faq.rst
```diff
@@ -60,6 +60,15 @@ the destructor (``__del__``) of browser. The solution is to
 call :func:`~mechanicalsoup.Browser.close` before the end of life of
 the object.
 
+Alternatively, you may also use the ``with`` statement which closes
+the browser for you::
+
+  def test_with():
+      with mechanicalsoup.StatefulBrowser() as browser:
+          browser.open(url)
+	  # ...
```
hemberger: Not important at all since this is just documentation, but this line has a tab in it, which makes it look funny in my 4-space tab setup.

tests/test_stateful_browser.py
```diff
@@ -295,6 +295,10 @@ def test_form_multiple():
     response = browser.submit_selected()
     browser.close()
 
+def test_with():
```
hemberger: I know it's not the job of documentation to teach people how python works, but I wouldn't have know that `with` calls `__enter__` and `__exit__` without looking it up. Maybe we could add a comment in the `test_with` that says that "testing with" <==> "testing enter and exit"? Certainly not necessary though!

moy: done.

PR #140, opened 2017-10-21
form.py: support <select multiple>

### LINKED ISSUE #124: Manage <select multiple>
`<select multiple>` should behave essentially like multiple checkboxes with the same name. For now, they don't: selecting one option deselects the previous ones.

@hemberger: while you're working on form, you may want to fix this too.

hemberger:
Fixes #124.

Normally, only one option of a select element can be selected.
But in a `<select multiple>`, it is valid HTML to select multiple
options.

We now support this with the following syntax:
```python
  # select option1 and option2
  form.set_select({'name': ('option1', 'option2')})
```

### DIFF
--- a/docs/ChangeLog.rst
+++ b/docs/ChangeLog.rst
@@ -40,6 +40,9 @@ Main changes:
   ``browser.get_current_form().print_summary()`` to get a summary of the
   fields you need to fill-in (and which ones are already filled-in).
 
+* The ``Form`` class now supports selecting multiple options in
+  a ``<select multiple>`` element.
+
 Bug fixes
 ---------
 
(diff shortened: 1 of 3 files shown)


hemberger:
Yep! This will always work since `Form.set` is now a direct wrapper around `Form.set_*`. (The only difference is that `set` can only take one name-value pair, but value can be as many elements as the `set_*` method allows.

I can certainly change the test from
```
form.set_select({'instrument': options)
```
to
```
browser['instrument'] = options
```
I only chose the former because it made it more explicit that it was testing `set_select`. Which would you prefer?

moy:
I tend to prefer calling `browser["name"]` in tests, to test the longest codepath. But we should also have a few tests that check directly the `set_*` methods, so let's merge the PR as-is.

PR #141, opened 2017-10-22
Re-enable bs4 warning and specify lxml as default parser

moy:


moy:
OK, added to the changelog and merged.

PR #142, opened 2017-10-22
Include tests directory in flake8 tests

hemberger:
This is mostly whitespace and line-length changes, as well as
a few unused variable fixes.

The tests `test_form_{noname,multiple}` now assert that the
request was successful.

----------------

Currently, the tests do not pass due to line-too-long flake8 errors (E501) in `test_{form,browser}.py`. The offending lines are the html samples. I'm not exactly sure how to handle these errors.

- [x] Fix E501 errors

### DIFF
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -73,7 +73,8 @@ choose_submit_form = '''
   <body>
     <!-- vaguely based on Trac edit-page form -->
     <form id="choose-submit-form" method="post" action="mock://form.com/post">
-      <textarea id="text" class="wikitext trac-resizable" name="text" cols="80" rows="40">
+      <textarea id="text" class="wikitext trac-resizable" name="text"
+        cols="80" rows="40">
       </textarea>
       <div class="field">
         <label>Comment about this change (optional):<br />
@@ -81,8 +82,8 @@ choose_submit_form = '''
         </label>
       </div>
       <div class="buttons">
-        <input type="submit" name="preview" value="Preview Page" /> 
-        <input type="submit" name="diff" value="Review Changes" /> 
+        <input type="submit" name="preview" value="Preview Page" />
+        <input type="submit" name="diff" value="Review Changes" />
         <input type="submit" id="save" name="save" value="Submit changes" />
         <button type="submit" name="cancel" value="Cancel" />
       </div>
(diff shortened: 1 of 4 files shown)


tests/test_form.py
```diff
@@ -212,11 +219,12 @@ def test_form_noaction():
 </html>
 '''
 
+
 def test_form_action():
     browser, url = setup_mock_browser()
-    browser.open_fake_page(submit_form_action, url="http://example.com/invalid/")
+    browser.open_fake_page(submit_form_action, url="http://fake.com/invalid/")
```
moy: example.com was really the right thing to do here. fake.com does exist, while example.com is specifically reserved for such usage: https://tools.ietf.org/html/rfc2606 / http://example.com/.

tests/test_stateful_browser.py
```diff
@@ -42,11 +46,11 @@ def test_submit_online():
     assert data["comments"] == "Some comment here"
     assert data["nosuchfield"] == "new value"
 
-    assert (json["headers"]["User-Agent"] ==
-            'testing https://github.com/hickford/MechanicalSoup')
+    assert (json["headers"]["User-Agent"] == 'testing MechanicalSoup')
```
moy: While you're there: unnecessary braces after assert (they were needed to allow the line break).

moy:
To solve remaining issues, at worse you have `# noqa:...` directives. http://flake8.pycqa.org/en/3.2.1/user/ignoring-errors.html

hemberger:
I tried using `# noqa:E501` at the end of the multi-line strings, but that didn't work for me (in fact, it only generated more warnings!). It's the individual lines of the multi-line string that flake8 is complaining about, so it seems like it would be weird -- and possibly wrong -- to have the noqa's there.

tests/test_form.py
hemberger: Huh! You learn something new every day! :D

moy:
I was thinking about a file-global `# flake8: noqa:501`, but it seems to disable all errors for the whole file (ie. back to where you started).

I also tried:
```

@@ -46,7 +47,7 @@ form_html = """
 <fieldset>
  <legend> Pizza Size </legend>
  <p><label> <input type=radio name=size value="small"> Small </label></p>
- <p><label> <input type=radio name=size value="medium" checked> Medium </label></p>
+ <p><label> <input type=radio name=size value="medium" checked> Medium </label></p> <!-- # noqa:E501 -->
  <p><label> <input type=radio name=size value="large"> Large </label></p>
 </fieldset>
 <fieldset>
```

Surprisingly, a single instance of this silents the warning for the whole file. Sounds like a flake8 bug.

Best is probably to actually cut the lines then.

PR #151, opened 2017-10-27
.travis-ci.yml: basic back-off strategy when httpbin is down

moy:
Hopefully this will reduce the failure rate on Travis-CI.

### DIFF
--- a/wait-for-httpbin.sh
+++ b/wait-for-httpbin.sh
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+# httpbin sometimes times out, but comes back shortly after this
+# happens. Apply a basic backoff scheme to give a chance to the tests
+# to run later instead of failing.
+for i in 1 2 3
+do
+    if curl http://httpbin.org >/dev/null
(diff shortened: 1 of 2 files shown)


wait-for-httpbin.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+# httpbin sometimes times out, but comes back shortly after this
+# happens. Apply a basic backoff scheme to give a chance to the tests
+# to run later instead of failing.
+for i in 1 2 3
+do
+    if curl http://httpbin.org >/dev/null
+    then
+	break
```
hemberger: Would be nice if you can change this to spaces instead of a tab, but it's not essential.

wait-for-httpbin.sh
```diff
@@ -0,0 +1,13 @@
...
+    fi
+    sleep $(($RANDOM % 100 + 100 * i))
```
hemberger: Should we print how long we're sleeping for?

PR #152, opened 2017-10-28
stateful_browser.py: add nr option to select_form

### LINKED ISSUE #130: Implement nr option?
In MechanicalSoup, almost all of our methods that find some element return or operate on the first one that is found (`Form.choose_submit` appears to be the recently-modified exception). If, for example, you have a page with two `<form>...</form>` with no unique attributes, then you cannot select the second form with `StatefulBrowser.select_form`.

This problem is solved in Mechanize by giving these methods an `nr=1` argument, which selects the _nr_-th match, and the first by default. We could certainly add this to many of the methods and retain the current default behavior entirely. I think this alone would be a worthwhile enhancement.

One related (but perhaps separate) issue is the lack of feedback when you omit the `nr` option and get an element that you don't expect, likely resulting in an indecipherable error, or worse, silently incorrect behavior. In my ideal world, you'd specify `nr` when you know you want the _nr_-th match, and if `nr` is omitted you demand that there is only one match. This seems like the safest option, but may be too draconian. Perhaps it would be suitable for a "debug" or "strict" mode?

hemberger:
Closes #130.

The `nr` option allows us to choose which form to select if
multiple forms match the `selector`. Defaults to the first
matching form.

In addition, `*args` and `**kwargs` are no longer forwarded
to BeautifulSoup's `.select`.

---------------

- [x] We should decide if we want `nr` to be a 0-based index (as it is currently coded, for consistency with mechanize) or to be a 1-based index (which might make the option a little easier to explain).

### DIFF
--- a/docs/ChangeLog.rst
+++ b/docs/ChangeLog.rst
@@ -23,7 +23,9 @@ Main changes:
 * We now have a documentation: https://mechanicalsoup.readthedocs.io/
 
 * ``StatefulBrowser.select_form`` can now be called without argument,
-  and defaults to ``"form"`` in this case.
+  and defaults to ``"form"`` in this case. It also has a new argument,
+  ``nr`` (defaults to 0), which can be used to specify the index of
+  the form to select if multiple forms match the selection criteria.
 
 * We now use requirement files. You can install the dependencies of
   Mechanicalsoup with e.g.::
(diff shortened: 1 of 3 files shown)


moy:
Looks good to me. 0-based index shouldn't be a problem for programmers.

PR #157, opened 2017-11-02
Prepare for next release

hemberger:
The next release is expected to be v1.0.0. While in development,
the version will be reported as "1.0.0-dev".

Since I cannot upload to PyPI, I'm making this available to be merged after the upload is finished, as per the release checklist.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -107,7 +107,7 @@ At each release:
 
 Right after the release:
 
-- Update the version number to a ``x.y.dev`` number in
+- Update the version number to a ``x.y.z-dev`` number in
   ``mechanicalsoup/__version__.py``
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.rst
```diff
@@ -107,7 +107,7 @@ At each release:
 
 Right after the release:
 
-- Update the version number to a ``x.y.dev`` number in
+- Update the version number to a ``x.y-dev`` number in
```
moy: The `.dev` notation is the recommandation for Python: https://www.python.org/dev/peps/pep-0440/#pre-releases

But semantic versionning (http://semver.org) requires `-dev`, and since PEP 440 allows `-` as an alias for `.`, why not go for it.

This should actually be x.y.z-dev.

Issue #158, opened 2017-11-04
Forms with repeated input elements


hemberger:
Here is a trimmed down version of the [Kickstarter Login page](https://www.kickstarter.com/login):
```html
<form>
  <input name="box" type="hidden" value="0"/>
  <input checked="checked" name="box" type="checkbox" value="1"/>
  <input type="submit" value="Submit" />
</form> 
```
A simple script to select and then submit this form results in the following error when using 88d39ea461f3c2210ece850e68696b6fb91df78c:
```
  File "mechanicalsoup/browser.py", line 191, in _prepare_request
    request = self._build_request(form, url, **kwargs)
  File "mechanicalsoup/browser.py", line 147, in _build_request
    data.setdefault(name, []).append(value)
  AttributeError: 'str' object has no attribute 'append'
```
The reason is because `_build_request` first parses the non-checkbox input, and sets the value with `data[name] = value`. Then it parses the checkbox input and tries to set the value with `data.setdefault(name, []).append(value)`, which assumes that any previous value for `name` was also a checkbox input and therefore is a list.

I'm not sure how the Kickstarter server is processing this input, but testing with [w3school's Try It](https://www.w3schools.com/html/tryit.asp?filename=tryhtml_default), it seems that the resulting QSL should be:
```
box=0&box=1 
```
However, since we store the input data in a dictionary, we tend to clobber repeated elements (except for checkboxes and submit multiple's). Perhaps we should instead keep all values of input elements with the same name, since that's what the browser would do.

(NOTE: if we make this change, I think `setup_mock_browser` should check `query` instead of `set(query)`.)

PR #159, opened 2017-11-04
Test full QSL in setup_mock_browser

hemberger:
Related to #158.

Compare the expected and posted QSL as lists rather than sets
to avoid clobbering multiple elements with the same name.

For example:
```python
parse_qsl('x=1&x=1') != parse_qsl('x=1')
```
even though
```python
set(parse_qsl('x=1&x=1')) == set(parse_qsl('x=1'))
```
Since valid HTML can produce both query strings, we should be
able to distinguish between them in our tests.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -44,7 +44,7 @@ def setup_mock_browser(expected_post=None, text=choose_submit_form):
         def text_callback(request, context):
             # Python 2's parse_qsl doesn't like None argument
             query = parse_qsl(request.text) if request.text else ()
-            assert(set(query) == set(expected_post))
+            assert list(query) == list(expected_post)
             return 'Success!'
         mock.register_uri('POST', url + '/post', text=text_callback)
 
--- a/tests/test_stateful_browser.py
+++ b/tests/test_stateful_browser.py
@@ -311,8 +311,8 @@ submit_form_multiple = '''
 
 
 def test_form_multiple():
-    browser, url = setup_mock_browser(expected_post=[('foo', 'tempeh'),
-                                                     ('foo', 'tofu')])
+    browser, url = setup_mock_browser(expected_post=[('foo', 'tofu'),
+                                                     ('foo', 'tempeh')])
     browser.open_fake_page(submit_form_multiple, url=url)
     browser.select_form('#choose-submit-form')
     response = browser.submit_selected()


tests/utils.py
```diff
@@ -44,7 +44,7 @@ def setup_mock_browser(expected_post=None, text=choose_submit_form):
         def text_callback(request, context):
             # Python 2's parse_qsl doesn't like None argument
             query = parse_qsl(request.text) if request.text else ()
-            assert(set(query) == set(expected_post))
+            assert list(query) == list(expected_post)
```
moy: Do we need this `list(...)`? If so, this deserves a comment to explain why.

moy:
OK, I think I understand the issue: we build `data` as a dictionary, which now is a dictionary of lists. Individual lists are ordered, but the dictionary isn't.

So, either we should

1) build data as a list of lists, or
2) use a comparison operator intermediate between "check the order" and "only check set": check number of repeated keys, but don't check the order of keys.

Option 1) might be better (in case some website relies on the order of keys), but AFAIK the order of fields in the data sent by browser is unspecified, so it's probably overkill to try to get the order right. So I'd lean towards 2), and this is rather simple to do: just sort the lists before comparison.

In short, I think we could write this instead:

    assert sorted(query) == sorted(expected_post)

tests/utils.py
hemberger: The `list` is needed to ensure that `query` and `expected_post` are the same type. I think `query` is always a list, but `expected_post` was sometimes a tuple. I can add a comment, but we could also consider removing the cast for `query`.

hemberger:
Yes, when checking the expected query string (at least with [TryIt](https://www.w3schools.com/html/tryit.asp?filename=tryhtml_form_submit)), the order was always the order of the input elements on the page. I don't know if this is set by any standard, but we're treading in the murky area of bad HTML design anyway, since repeated element names is rarely a good idea. :)

moy:
The non-deterministic ordering we get is not for repeated elements, but for different keys i.e. generating `x=1&y=2` or `y=2&x=1`.

hemberger:
I see, so for different keys, I can't imagine that the ordering matters (as you were saying), but if we use `sorted` to check the QSL, then won't that hide any errors with the ordering of elements with the same name?

moy:
Then, we can use a sort only on the keys and keep the values ordered as they were:

```
def c(x, y):
    return cmp(x[0], y[0])
sorted(x, c)
```

hemberger:
I think that if we refactor `_build_request` (something I was thinking about today, and may take a stab at tomorrow), then we can pass Requests a list of tuples for `data`. Looking at its source code, the list of tuples will create a deterministic query string:
```python
    def _encode_params(data):
        """Encode parameters in a piece of data.

        Will successfully encode parameters when passed as a dict or a list of
        2-tuples. Order is retained if data is a list of 2-tuples but arbitrary
        if parameters are supplied as a dict.
        """
```
Then we can do a simple equality operator and test that ordering is retained.

moy:
Yes, that would work, but that just makes the code a bit heavier on our side:

First, you can't `data[key] = value` anymore. But a straightforward helper function should allow us to deal with a list of tuples while retaining a clean code on our side.

But more importantly, if we want the order to be the text order, we can't encode data as we do currently by iterating through elements types by types. Currently, we process all `input` tags, then all `textarea`, and then all `select`.

moy:
Superseded by #219 .

PR #160, opened 2017-11-04
Fix QSL for repeated form elements

### LINKED ISSUE #158: Forms with repeated input elements

hemberger:
Closes #158.

Improve consistency of query string construction between
MechanicalSoup and web browsers in edge cases where form elements
have duplicate *name*-attributes.

This change affects duplicate-named `input` elements that don't
have a *type*-attribute and `textarea` elements; all values will
show up in the query string instead of just the last.

For example, if a form with the following is submitted:
```
<input name="key" value="val1" />
<input name="key" value="val2" />
```
Then the resulting query string will now be
```
key=val1&key=val2
```
instead of
```
key=val2
```
This also fixes an even more unlikely scenario where an exception
is thrown when a checkbox and a type-less input have the same name.

### DIFF
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -143,9 +143,9 @@ class Browser(object):
                 # web browsers use empty string for inputs with missing values
                 value = input.get("value", "")
 
-            if input.get("type") == "checkbox":
-                data.setdefault(name, []).append(value)
-
+            if input.get("type") == "radio":
+                # Uses the last checked radio (only one should be checked)
+                data[name] = value
             elif input.get("type") == "file":
                 # read http://www.cs.tut.fi/~jkorpela/forms/file.html
                 # in web browsers, file upload only happens if the form"s (or
@@ -156,15 +156,14 @@ class Browser(object):
                 if isinstance(value, string_types):
                     value = open(value, "rb")
                 files[name] = value
-
             else:
-                data[name] = value
+                data.setdefault(name, []).append(value)
 
         for textarea in form.select("textarea"):
             name = textarea.get("name")
             if not name:
                 continue
-            data[name] = textarea.text
+            data.setdefault(name, []).append(textarea.text)
 
         for select in form.select("select"):
             name = select.get("name")
--- a/docs/ChangeLog.rst
+++ b/docs/ChangeLog.rst
@@ -5,6 +5,16 @@ Release Notes
 Version 1.0 (in development)
 ============================
 
+Bug fixes:
+----------
+
+* Improve consistency of query string construction between MechanicalSoup
+  and web browsers in edge cases where form elements have duplicate names.
+  This only affects duplicate-named ``input`` elements that don't have a
+  *type* and ``textarea`` elements; all values will show up in the query
+  string instead of just the last.
+  See `#158 <https://github.com/MechanicalSoup/MechanicalSoup/issues/158>`__.
+
 Version 0.9
 ===========
 
--- a/tests/test_browser.py
+++ b/tests/test_browser.py
@@ -70,9 +70,9 @@ def test_build_request():
     browser = mechanicalsoup.Browser()
     request = browser._build_request(form)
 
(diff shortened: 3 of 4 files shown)


mechanicalsoup/browser.py
```diff
@@ -143,9 +143,9 @@ def _build_request(self, form, url=None, **kwargs):
                 # web browsers use empty string for inputs with missing values
                 value = input.get("value", "")
 
-            if input.get("type") == "checkbox":
-                data.setdefault(name, []).append(value)
-
+            if input.get("type") == "radio":
+                # Uses the last checked radio (only one should be checked)
+                data[name] = value
```
moy: Shouldn't this be `[value]` instead of just `value`?

If you have one `<input type="radio">` and another input, the code is later going to attempt a `.append` on this value and will crash.

hemberger: Yes, we could do that to avoid errors similar to #158. However, if we're going to go that route, then we need to make even more changes. For example,
```html
<form action="/action_page.php">
  <input name="box" value="0"/>
  <input checked="checked" name="box" type="radio" value="1"/>
  <input type="submit" value="Submit" />
</form>
```
This should result in a `box=0&box=1` query string (setting aside the fact that this seems like a bad HTML design decision). However, with just `data[name] = [value]`, it will clobber the 0 value from the type-less input, giving only `box=1`.

If we were to change the behavior for type=radio to `data.setdefault(name, []).append(value)` like the others, then this HTML:
```html
<form action="/action_page.php">
  <input name="box" type="radio" value="0"/>
  <input checked="checked" name="box" type="radio" value="1"/>
  <input type="submit" value="Submit" />
</form>
```
would result in `box=0&box=1`, even though it should only be `box=1`.

Off the top of my head, I can't think of a way to retain the correct single-valued results for types that are only allowed to have one value, while still accounting for the possibility of poorly designed HTML like the above.

moy: Perhaps we need another dict `data_radio` that we can fill-in with plain `data_radio[key] = value`, and once we're done iterating through the form, copy from `data_radio` to `data` using `setdefault(..., []).append()`?

hemberger: We could do that. Are there any other elements that are forcibly unique like `radio` that we would need to account for as well?

I worry about this function (which is already not organized in the most legible way) becoming too complex.

moy: At some point, we'll need some refactoring (perhaps extracting this "form -> data" to a separate file/class) I guess. I never liked this function ;-).

hemberger:
Superseded by #219

PR #162, opened 2017-11-06
.travis.yml: enable the pip cache

hemberger:
This caches `$HOME/.cache/pip` in Travis builds.

Our Travis build time is often dominated by dependency installation,
so caching pip can speed up builds significantly.

The cache can be managed at:
https://travis-ci.org/MechanicalSoup/MechanicalSoup/caches

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@
 
 sudo: false
 language: python
+cache: pip
 python:
   - 2.7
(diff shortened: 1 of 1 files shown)


moy:
I tried restarting a build and it does use the cache. The timing gain seems to be minimal, but the cost on our side is negligible, so let's do this.

PR #164, opened 2017-11-07
Replace httpbin.org with pytest-httpbin in tests

hemberger:
The pytest-httpbin module provides pytest support for the
httpbin module (which is the code that runs the remote server
http://httpbin.org). This locally spins up an internal webserver
when tests are run.

With this change, MechanicalSoup tests can be run without an
internet connection. As a result, the tests run much faster.

You may need the python{,3}-dev package on your system to pip
install the pytest-httpbin module.


hemberger:
A note about the pypy3 build failure:

The last few lines of the (very long) exception are:
```
 File "/home/travis/virtualenv/pypy3.5-5.8.0/site-packages/dateparser/__init__.py", line 4, in <module>
    from .date import DateDataParser
  File "/home/travis/virtualenv/pypy3.5-5.8.0/site-packages/dateparser/date.py", line 8, in <module>
    import regex as re
  File "/home/travis/virtualenv/pypy3.5-5.8.0/site-packages/regex.py", line 398, in <module>
    import _regex_core
  File "/home/travis/virtualenv/pypy3.5-5.8.0/site-packages/_regex_core.py", line 21, in <module>
    import _regex
ImportError: /home/travis/virtualenv/pypy3.5-5.8.0/site-packages/_regex.pypy3-58-x86_64-linux-gnu.so: undefined symbol: PyUnicode_FromKindAndData
```

I had a look around and found some similar issues with `regex` and `pypy`:
* https://bitbucket.org/mrabarnett/mrab-regex/issues/240/unable-to-build-the-project-from-source-on
* https://bitbucket.org/mrabarnett/mrab-regex/issues/253/run-into-error-under-pypy-580

but I honestly can't tell if this is fixed and some module along the chain hasn't updated to a version that includes the bugfix, or if it's still an open issue.

I would really like to use this PR, because it makes the tests run 5x faster for me, but we may need to work our way along the import chain until this is fixed. :(

moy:
Is it possible/easy to keep the tests as they are for pypy3 and use pytest-httpbin for other builds?

Does adding `--upgrade` to all the `pip install` commands help?

hemberger:
I tried adding `--upgrade` to the install commands, but that didn't help. I would rather not have both implementations in place if we can help it (I think that would get a bit complicated), and I'm willing to keep this PR on hold while I investigate further, but I'm not even sure where to start. Should I start by creating an issue for pytest-httpbin, or perhaps go straight to regex?

moy:
First thing I'd do is to reproduce the issue on a minimalistic example (no MechanicalSoup, just a single `.py` file that imports the guilty package (regex if possible, pytest-httpbin if regex alone doesn't crash), as much as possible within a docker image, or at least a Python virtualenv.

hemberger:
With an unimaginably simple Dockerfile, I can reproduce this error with _only_ `pytest-httpbin` in pypy3-5.8.0 (which is the version of pypy3 that Travis uses by default).

```Dockerfile
FROM pypy:3-5.8.0
#FROM pypy:3-5.9.0

RUN pip install pytest-httpbin
RUN pip install pytest

CMD ["pytest", "-h"]
```
Some probably good news: this error is _not_ encountered with pypy3-5.9.0. I'm not sure if we would be waiting on virtualenv or Travis (or both) to provide a 5.9.0 environment.

hemberger:
To clarify, this is an issue with `regex` alone.

```Dockerfile
FROM pypy:3-5.8.0
#FROM pypy:3-5.9.0

RUN pip install regex

ENTRYPOINT ["pypy3"]
```
with `-c "import regex"` passed to `docker run` gives the expected error:
```
  File "<string>", line 1, in <module>
  File "/usr/local/site-packages/regex.py", line 398, in <module>
    import _regex_core
  File "/usr/local/site-packages/_regex_core.py", line 21, in <module>
    import _regex
ImportError: /usr/local/site-packages/_regex.pypy3-58-x86_64-linux-gnu.so: undefined symbol: PyUnicode_FromKindAndData
```

hemberger:
I've submitted a bug report with the regex module. We'll see where that takes us.

https://bitbucket.org/mrabarnett/mrab-regex/issues/264/failure-to-import-regex-in-pypy3-580

hemberger:
The issue was too extensive for the maintainer of `regex` to fix, so we will need to wait until Travis-CI allows using pypy3-5.9.0 before we can consider merging this.

moy:
It's actually tempting to disable the pypy3 build until this is repaired. We have python3 and pypy build, so we're already 99.9% sure we won't break pypy3, and it'll still be possible to run pypy3 tests manually before the next release.

hemberger:
It's up to you! I've really appreciated using this commit locally, but it's hard for me to weigh whether it's worth limiting our tests.

I'm guessing we shouldn't expect a useful response (if we get one at all) from Travis if we asked when pypy3-5.9.0 will be added..

ghisvail:
> It's up to you! I've really appreciated using this commit locally, but it's hard for me to weigh whether it's worth limiting our tests.

This commit could allow Debian to run the tests for the corresponding packages (`python-mechanicalsoup` and `python3-mechanicalsoup`). Debian builders prevent access to the network at build time to avoid unattended tempering of the builds.

hemberger:
Unfortunately it appears that Travis is still using pypy3-5.8.0. As suggested in https://github.com/MechanicalSoup/MechanicalSoup/pull/164#issuecomment-345016081, we can disable the pypy3 tests on Travis if we feel that this PR is more important than the automatic tests. Alternatively, we could try to encourage Travis to update their pypy3 environment.

hemberger:
This is closed by #209, which uses the new naming scheme for Travis' pypy environments to get a newer pypy3 version.

PR #166, opened 2017-11-10
browser.py: refactor Browser.submit method

### LINKED ISSUE #165: Unexpected keyword argument 'proxies'
The use of requests.Request base class does not include a proxies keyword like the requests.request class.

```python
import requests
import mechanicalsoup
self.Session = requests.Session()
self.browser = mechanicalsoup.StatefulBrowser(session=self.Session)
self.browser.get(searchURL.url, proxies=self.proxy)
```

Full error log is:

```
    response = self.browser.submit_selected(proxies=self.proxy)
  File "C:\Users\Kevin\Desktop\Project\dev\lib\site-packages\mechanicalsoup\stateful_browser.py", line 194, in submit_selected
    *args, **kwargs)
  File "C:\Users\Kevin\Desktop\Project\dev\lib\site-packages\mechanicalsoup\browser.py", line 209, in submit
    request = self._prepare_request(form, url, **kwargs)
  File "C:\Users\Kevin\Desktop\Project\dev\lib\site-packages\mechanicalsoup\browser.py", line 191, in _prepare_request
    request = self._build_request(form, url, **kwargs)
  File "C:\Users\Kevin\Desktop\Project\dev\lib\site-packages\mechanicalsoup\browser.py", line 188, in _build_request
    return requests.Request(method, url, files=files, **kwargs)
TypeError: __init__() got an unexpected keyword argument 'proxies'
```

hemberger:
Closes #165.

Since both `Browser._prepare_request` and `Browser._build_request` are
internal methods, we can replace them with a single `Browser._request`.

This new method is the same as `_build_request`, except that it calls
`self.Session.request` to prepare and send the request instead of doing
it in two separate steps.

Since `requests.Session.request` both prepares and sends the request,
it has an API that includes keywords for both actions. The keywords
are a superset of `requests.Request` (except for `hooks`, but since
MechanicalSoup did not expose the resulting Request object, I believe
the hooks could not be used).

As a result, the `kwargs` that can be passed to `Browser.submit` has
expanded to include keywords related to sending the request.

### DIFF
--- a/docs/ChangeLog.rst
+++ b/docs/ChangeLog.rst
@@ -5,6 +5,19 @@ Release Notes
 Version 1.0 (in development)
 ============================
 
+Main changes:
+-------------
+* ``Browser.submit`` and ``StatefulBrowser.submit_selected`` accept a larger
+  number of keyword arguments. Arguments are forwarded to
+  `requests.Session.request <http://docs.python-requests.org/en/master/api/#requests.Session.request>`__.
+  [`#166 <https://github.com/MechanicalSoup/MechanicalSoup/pull/166>`__]
+
+Internal changes:
+-----------------
+* Private methods ``Browser._build_request`` and ``Browser._prepare_request``
+  have been replaced by a single method ``Browser._request``.
+  [`#166 <https://github.com/MechanicalSoup/MechanicalSoup/pull/166>`__]
+
 Version 0.9
 ===========
 
--- a/tests/test_browser.py
+++ b/tests/test_browser.py
@@ -64,26 +64,26 @@ form_html = """
 """
 
 
-def test_build_request():
+def test__request():
     form = BeautifulSoup(form_html, "lxml").form
 
     browser = mechanicalsoup.Browser()
-    request = browser._build_request(form)
+    response = browser._request(form)
 
-    assert request.data["customer"] == "Philip J. Fry"
-    assert request.data["telephone"] == "555"
-    assert request.data["comments"] == "freezer"
-    assert request.data["size"] == "medium"
-    assert request.data["topping"] == ["cheese", "onion"]
-    assert request.data["shape"] == "square"
+    data = response.json()['form']
+    assert data["customer"] == "Philip J. Fry"
+    assert data["telephone"] == "555"
+    assert data["comments"] == "freezer"
+    assert data["size"] == "medium"
+    assert data["topping"] == ["cheese", "onion"]
+    assert data["shape"] == "square"
 
-    request = browser._prepare_request(form)
-    assert "application/x-www-form-urlencoded" in request.headers[
+    assert "application/x-www-form-urlencoded" in response.request.headers[
         "Content-Type"]
     browser.close()
 
 
-def test_prepare_request_file():
+def test__request_file():
     form = BeautifulSoup(form_html, "lxml").form
 
     # create a temporary file for testing file upload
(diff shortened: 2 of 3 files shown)


tests/test_browser.py
```diff
@@ -94,15 +94,18 @@ def test_prepare_request_file():
...
+    response = browser._request(form)
+
+    # Check the response only "files" includes a "pic" keyword
+    data = response.json()
+    for key in data.keys():
+        if key == "files":
+            assert data[key]["pic"] == ":-)"
+            continue
+        with pytest.raises(Exception):
+            data[key]["pic"]
```
moy: The 2 lines above sound like a complex way to write

    assert "pic" not in data[key]

moy:
Don't forget to update the changelog.

I'm wondering whether we're loosing some potential flexibility by doing this. Currently, we're not, but what if we ever want to provide a hook that the user could execute between the "prepare" and the "send" step? Will it still be possible?

tests/test_browser.py
hemberger: I wish it was, but because `data[key]` can be both a dict and `None`, you get both `IndexError` and `TypeError` when you try to use an iterator (e.g. `in`) with it.

hemberger:
ChangeLog has been updated.

Your question is a good one. From the current version of the code, we're only gaining generality. Since request construction and sending was all done internally in `submit`, even without this PR we would need to change the API to allow users to modify the prepared request before sending. I don't think this should be any harder of a task after this PR is merged. However, it might never be unnecessary -- I think the idea behind `requests.Sessions.request` is to allow people to specify almost anything you'd need to do for both the preparing and sending in a single call.

tests/test_browser.py
moy: OK, but then I'd rather make this explicit in the code with stg like `assert data[key] is None or "pic" not in data[key]`. That would make it clear to the reader what the possible values are, and would avoid this `with pytest.raises(Exception):` : not really dangerous here, but it's a construct in which any buggy code that raise an exception makes the test pass...

hemberger: Good point! How does it look now?

moy: Good.

moy:
> without this PR we would need to change the API to allow users to modify the prepared request before sending

Yes, we would need to modify our API, but it's rather simple to provide a `add_hook(callback)` and to call the callback between prepare and send before the PR.

> I don't think this should be any harder of a task after this PR is merged.

The problem is that the lines of code between prepare and send are currently in our code, so we can still add whatever we want (like calling a user-provided callback) there. After the PR, the only lines of code between prepare and send are in the requests library, we can't modify them anymore.

>  I think the idea behind requests.Sessions.request is to allow people to specify almost anything

Yes, but the question is whether this "almost" is sufficient for 100% use-cases. The request library provides the flexibility of allowing tweaking the prepared request before sending:

http://docs.python-requests.org/en/master/user/advanced/#prepared-requests

We don't currently need it, but it'd be nice to keep the possibility to do it in the future.

Re-reading the code, I think there's another way to expose the full flexibility of the requests library: merge your PR and then split `_request` into:

* `_extract_form_data()` that would build `kwargs` (i.e. the current `_request()`, except its last line)
* `_send_form_data()` that would call `self.session.request(..., kwargs)`

(Or perhaps these would deserve a name not starting with `_`)

and expose `_request()` that would just call these two functions. This way, the user may, if needed, call `_extract_form_data()` and do whatever tweaking is needed to actually send the HTTP request. This would slightly increase the flexibility of the code, and would also split the form-related code and the http-related code better.

I short, I think we can merge the PR, and we may want to do the split I suggest later.

docs/ChangeLog.rst
```diff
@@ -5,6 +5,19 @@ Release Notes
...
+* ``Browser.submit`` and ``StatefulBrowser.submit_selected`` accept a larger
+  number of keyword arguments. Arguments are forwarded to
+  `requests.Session.request <http://docs.python-requests.org/en/master/api/#requests.Session.request>`__.
+  [`#166 <https://github.com/MechanicalSoup/MechanicalSoup/pull/166>`__]
+
+Internal changes:
+-----------------
+* Private methods ``Browser._build_request`` and ``Browser._prepare_request``
+  have been replaced by a single method ``Browser._request``.
+  [`#166 <https://github.com/MechanicalSoup/MechanicalSoup/pull/166>`__]
```
moy: I usually omit these internal changes from the ChangeLog. I've added "Internal changes" section for other releases for things we should be proud of like 100% coverage and that didn't fit elsewhere, or stuff that developers should be aware of like changes in the way to launch testsuite. But it doesn't harm to have them.

hemberger:
Yep, that split sounds completely natural to me. The problem with the previous split was that it combined `prepare` and `send`, but had no way of conveniently passing any parameters to `send`.

docs/ChangeLog.rst
hemberger: The only reason I added these was because there is no truly private method in python, so someone could have been using them. I have no problem with this item being omitted from the ChangeLog. Please feel free to do so if that's your preference!

PR #171, opened 2017-12-07
Fixed a ReferenceError when using Python 3.4.3

Harapuia:
This is a fix for ReferenceError: weakly-referenced object no longer exists when mechanicalsoup is used on Python 3.4.3. self.session may have already been garbage collected by the time browser.py gets deleted, so calling self.session.close() will result in this error.

It's resolved by using weakref.finalize instead

### DIFF
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -38,7 +38,12 @@ class Browser(object):
         self.__raise_on_404 = raise_on_404
         self.session = session or requests.Session()
 
-        self._finalize = weakref.finalize(self.session, self.close)
+        if hasattr(weakref, 'finalize'):
+            self._finalize = weakref.finalize(self.session, self.close)
+        else:
+            # Python < 3 does not have weakref.finalize, but these
+            # versions accept calling session.close() within __del__
+            self._finalize = self.close
 
         self.set_user_agent(user_agent)
 
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -7,6 +7,7 @@ import webbrowser
 import tempfile
 from .utils import LinkNotFoundError
 from .__version__ import __version__, __title__
+import weakref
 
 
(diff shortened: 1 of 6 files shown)


mechanicalsoup/browser.py
```diff
@@ -37,6 +38,8 @@ def __init__(self, session=None, soup_config={'features': 'lxml'},
         self.__raise_on_404 = raise_on_404
         self.session = session or requests.Session()
 
+        self._finalize =  weakref.finalize(self.session,self.close)
```
moy: Nit: missing space after `,`. I guess flake8 will tell you when the Travis-CI build completes ;-).

moy: Oh, and extra-space after `=`.

Harapuia:
I've made the changes as requested

I've no idea why it happens, but when i investigated the issue that i had with it the official advice was to do it like this, instead of using __del__. So i decided to change it properly here

Now to figure out where those attribute errors are coming from..
i don't think i'm doing anything nuts

moy:
OK, I've pushed 2 commits on top of yours (the first is needed for the Travis-CI build to pass on Python 2.7). If you're OK with this, I'll squash them together and merge.

Thanks!

Harapuia:
I'm very okay with this. I thought it was a clever fix as i was already fumbling about with weakref.ref() trying to get a replacement for 2.7 to work. This was much smarter!

Glad i could help!

hemberger:
Thanks for the contribution @Harapuia, this looks great! It seems odd to me that `self._finalize` could be constructed without error when `weakref.finalize` doesn't exist, but I'm glad that `self.close()` works in these versions!

Harapuia:
It can be constructed because self._finalize is just a variable name. Maybe it should be renamed for additional clarity?

moy:
Merged (manually to squash everything together, and I took the liberty of amending the commit message).

PR #172, opened 2017-12-07
stateful_browser.py: add internal _BrowserState class

hemberger:
This is a simple data structure that holds the page, url, and form.
It makes managing the state of the StatefulBrowser much easier.

-------------

I'm doing some work on adding a `download` method, and having an easier way to manage the state of the browser is a useful prerequisite.

### DIFF
--- a/mechanicalsoup/stateful_browser.py
+++ b/mechanicalsoup/stateful_browser.py
@@ -9,6 +9,13 @@ import re
 import bs4
 
 
+class _BrowserState:
+    def __init__(self, page=None, url=None, form=None):
+        self.page = page
+        self.url = url
+        self.form = form
+
+
 class StatefulBrowser(Browser):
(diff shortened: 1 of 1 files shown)


mechanicalsoup/stateful_browser.py
```diff
@@ -138,10 +141,8 @@ def open_fake_page(self, page_text, url=None, soup_config=None):
         URL. Useful mainly for testing.
         """
         soup_config = soup_config or self.soup_config
-        self.__current_page = bs4.BeautifulSoup(
-            page_text, **soup_config)
-        self.__current_url = url
-        self.__current_form = None
+        self.__state = _BrowserState(
+            page=bs4.BeautifulSoup(page_text, **soup_config), url=url)
```
moy: Style: I'd add a line break before `url=`.

PR #173, opened 2017-12-07
Add pytest fixture to prepare for pytest-httpbin

hemberger:
We add a testing utils class `HttpbinWrapper`, which is simply
a drop-in replacement for the `httpbin` fixture that pytest-httpbin
provides.

This will allow us to write tests in the style of pytest-httpbin
until we can merge PR #164 into master.

### DIFF
--- a/tests/test_browser.py
+++ b/tests/test_browser.py
@@ -7,10 +7,10 @@ from requests.cookies import RequestsCookieJar
 import pytest
 
 
-def test_submit_online():
+def test_submit_online(httpbin):
     """Complete and submit the pizza form at http://httpbin.org/forms/post """
     browser = mechanicalsoup.Browser()
-    page = browser.get("http://httpbin.org/forms/post")
+    page = browser.get(httpbin + "/forms/post")
     form = page.soup.form
 
     form.find("input", {"name": "custname"})["value"] = "Philip J. Fry"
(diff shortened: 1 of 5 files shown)


tests/test_browser.py
```diff
@@ -7,10 +7,16 @@
 import pytest
 
 
-def test_submit_online():
+@pytest.fixture
+def httpbin():
+    from utils import HttpbinWrapper
+    return HttpbinWrapper()
```
moy: Ideally the 3 instances of this would be factored into a common file. If I read correctly https://docs.pytest.org/en/latest/fixture.html#conftest-py-sharing-fixture-functions, moving it to a `conftest.py` file should do it.

But since this is meant to be temporary anyway, I don't have strong objection in keeping it as-is.

moy: Oh, and could this become a conditional code that would actually do the pytest-httpbin thing when it works, and fall back to HttpbinWrapper if not?

hemberger:
This is my latest best attempt, where I use `conftest.py` as you suggest.

I tried a few things to make this fixture conditional on `pytest_httpbin` not being installed. The most obvious and simplest was to replace the contents of `conftest.py` with:
```python
try:
    import pytest_httpbin
except ImportError:
    @pytest.fixture
    def httpbin():
        from utils import HttpbinRemote
        return HttpbinRemote()
```
Unfortunately this gives a pretty messy error culminating in the line
```
INTERNALERROR> _pytest.vendored_packages.pluggy.PluginValidationError: unknown hook 'pytest_httpbin' in plugin <module 'conftest' (<_pytest.assertion.rewrite.AssertionRewritingHook object at 0x7f0b4d92ee10>)>
```

However, if I move this code back to where it originally was at the top of each `test_*.py` file, then this try-except trick works (apart from the warning about `pytest_httpbin` being unused -- that's probably an easy fix though). Or this could be moved into a free function in `utils.py` that we call in each of these files.

But perhaps the least intrusive option is to keep it as is now, with no automatic detection of `pytest_httpbin`. And then if we want to use `pytest_httpbin`, we simply delete `conftest.py`. Thoughts?

moy:
Let's merge this. It'll still be time to improve it later if we get better ideas.

moy:
Indeed it's harder than it seems. My best try was:

<pre>
@pytest.fixture(scope='session')
def httpbin(request):
    try:
        import pytest_httpbin
        return pytest_httpbin.plugin.httpbin(request)
    except ImportError:
        from utils import HttpbinRemote
        return HttpbinRemote()
</pre>

It kind of works, but it actually doesn't solve our issue: if pytest-httpbin is installed, then the code crashes before reaching this point, since pytest does the `import pytest_httpbin` before we get a chance to execute our code.

Unfortunately, just deleting `conftest.py` is not enough, one also has to edit `tests/requirements.txt` to get the real `pytest_httpbin` fixture with `python setup.py test`.

If we want this conditional activation, the best idea I came up with would be to adapt `tests/requirements.txt` and delete `conftest.py` before launching the tests in `.travis-ci.yml`. Probably overkill, so I'll stop here.

PR #182, opened 2018-01-02
stateful_browser.py: add download_link method

### LINKED ISSUE #170: [How-to]: is there a "download" function?
Hi,
first of all thank you, this is a great tool!!

I have writter a stupid hello world example (below). At the end I have an URL for which I would like to start a file download.

How to do it?

Thank you

```python
import mechanicalsoup
browser.open("http://albopretorio.comune.palermo.it/albopretorio/jsp/home.jsp?modo=info&info=servizi.jsp&ARECOD=70&SERCOD=-1&sportello=albopretorio")
browser.follow_link("TD=60")
browser.follow_link("TD=2013")
browser.follow_link("row=0")
myPdfURL=browser.follow_link("idx=0")
myPdfURL.url
```

hemberger:
Closes #170.

This new method will download the contents of a link to a specified
file without changing the state of the browser (unlike `follow_link`
and similar methods).

--------------

I was not completely certain about the method name, implementation, and how to best test it, so if there are any suggestions, I'm open to further discussion. This is fully functional as far as I'm aware, and ready to be merged, but please consider it a first design draft.

### DIFF
--- a/mechanicalsoup/stateful_browser.py
+++ b/mechanicalsoup/stateful_browser.py
@@ -278,16 +278,14 @@ class StatefulBrowser(Browser):
             These specify a link to the page whose contents you want to
             download.
         """
-        # Store the current browser state
-        previous_state = self.__state
-        resp = self.follow_link(*args, **kwargs)
-
-        # Restore the previous browser state
-        self.__state = previous_state
+        url = self.follow_link(return_url_only=True, *args, **kwargs)
+        response = self.session.get(url)
+        if self.raise_on_404 and response.status_code == 404:
+            raise LinkNotFoundError()
 
         # Save the response content to file
         with open(filename, 'wb') as f:
-            f.write(resp.content)
+            f.write(response.content)
 
     def launch_browser(self, soup=None):
         """Launch a browser to display a page, for debugging purposes.
--- a/mechanicalsoup/stateful_browser.py
+++ b/mechanicalsoup/stateful_browser.py
@@ -264,6 +264,27 @@ class StatefulBrowser(Browser):
                 raise
         return self.open_relative(link['href'])
 
+    def download_link(self, filename, *args, **kwargs):
+        """Downloads the contents of a link to a file. The browser state
+        will not change when calling this function.
+
+        :param filename: Filesystem path where the page contents will be
+            downloaded.
+        :param \*args, \*\*kwargs: Arguments forwarded to :func:`follow_link`.
+            These specify a link to the page whose contents you want to
+            download.
+        """
+        # Store the current browser state
+        previous_state = self.__state
+        resp = self.follow_link(*args, **kwargs)
+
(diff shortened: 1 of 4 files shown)


moy:
I don't like the "save and restore state" approach. In general, if you don't want to modify something, it's best to leave it untouched that to save+restore it. For example, if you get an exception during the call to `self.follow_link` in your code, you'll skip the restore part and get the browser in an inconsistent state (this can happen if you follow a link to a 404 error).

I've pushed 3 commits on top of yours (we should merge my commit #3 with yours if we go this way) to implement the same feature without save/restore. What do you think?

moy:
About the method name, I don't like `follow_link` much because it doesn't download a "link", but its destination. However, I'm not sure I can propose a better alternative: it is about a link since it shares the same API as follow_link. Perhaps we can just do the job within `follow_link` by doing something like

    def follow_link(self, ..., download_to=None, ...):
        ...
        if download_to is not None:
            # Save response content to file 'download_to'
        else:
            # Current behavior of follow_link

mechanicalsoup/stateful_browser.py
```diff
@@ -237,7 +237,7 @@ def find_link(self, *args, **kwargs):
         else:
             return links[0]
 
-    def follow_link(self, link=None, *args, **kwargs):
+    def follow_link(self, link=None, return_url_only=False, *args, **kwargs):
```
hemberger: I think it would be confusing for `follow_link` to have an option that makes it not follow a link and instead do something that seems more like a behavior of `find_link`.

hemberger:
Do you mean you don't like `download_link` as the name? If so, I agree it's not the best, but I think it clearly conveys to the user what it does. One alternative is to adapt the right-click "Save link as..." wording that all browsers seem to use as the option to download the contents of a link.

The reason I prefer using a separate method is because I think it would be unintuitive to have an argument that completely changes the behavior of `follow_link`. Of course there will always be arguments in any code that can only be understood by studying the documentation, but I bet we'll get a lot more "How do I download a link?" questions if the option is tucked away as an argument to `follow_link` than if it's its own method.

hemberger:
To answer your question, while save+restore state is simple and easy to maintain, I think your arguments against it are sound, and the implementation you've offered is more technically correct. I'm happy to go with yours!

moy:
> Do you mean you don't like download_link as the name?

Just that it suggests that we "download a link", while we actually follow the link to download the target page.

But I agree with your arguments, and I'd add that someone looking for a way to download a file is likely to grep "download" in your API, hence the name should be `download_<something>`. So `download_link` is not that bad. Perhaps `download_link_target` would be more technically accurate, but it's probably overkill.

In short: let's go with `download_link`. I'll add the missing tests and rebase+squash commits soon.

moy:
Let's continue the discussion on #184 .

PR #178, opened 2017-12-19
Sanitize argument forwarding between "follow_link" and "links"

amotl:
Hi Dan,

this is a quick attempt to address #177 we just talked about.

With kind regards,
Andreas.

Issue #177, opened 2017-12-19
Problems calling "follow_link" with "url_regex"


amotl:
Dear Dan and Matthieu,

first things first: Thanks for conceiving and maintaining this great library. We also switched one of our implementations over from "mechanize" as per https://github.com/ip-tools/ip-navigator/commit/a26c3a8a and it worked really well.

When doing so, we encountered a minor problem when trying to call the `follow_link` method with the `url_regex` keyword argument like
```
response = self.browser.follow_link(url_regex='register/PAT_.*VIEW=pdf', headers={'Referer': result.url})
```

This raises the exception
```
TypeError: links() got multiple values for keyword argument 'url_regex'
```

I am are currently a bit short on time, otherwise i would have submitted a pull request without further ado. Thanks a bunch for looking into this issue.

With kind regards,
Andreas.

hemberger:
Hi Andreas,

Thanks for the kind words! Though our thanks must be extended to M. Hickford who was the original author of this library.

From the documentation of [StatefulBrowser.follow_link](https://mechanicalsoup.readthedocs.io/en/stable/mechanicalsoup.html#mechanicalsoup.StatefulBrowser.follow_link),
> If ``link`` doesn't have a *href*-attribute or is None, treat
> ``link`` as a url_regex and look it up with :func:`find_link`.
> Any additional arguments specified are forwarded to this function.

I think this is a bit unclear, but what it really means is this:

If the `link` argument given to `follow_link` is not a bs4 tag, then pass it as the `url_regex` argument to `find_link`.

Therefore, in your call to `follow_link`:
```
response = self.browser.follow_link(url_regex='register/PAT_.*VIEW=pdf', headers={'Referer': result.url})
```
you're essentially making this call to `find_link` (since `link=None`):
```
response = self.browser.find_link(url_regex=None, url_regex='register/PAT_.*VIEW=pdf', headers={'Referer': result.url})
```
To fix your error, you can simply omit `url_regex=` or change it to `link=` in your call to `follow_link`:
```
response = self.browser.follow_link('register/PAT_.*VIEW=pdf', headers={'Referer': result.url})
```

I'm not particularly fond of how all the argument forwarding is set up, and it certainly makes it more difficult to use correctly. At the very least we can be more explicit in the documentation, but perhaps there's also some way to refactor these methods without impacting backwards compatibility too much.

Please let us know if this resolves your problem!

amotl:
Dear Dan,

[...]

PR #184, opened 2018-01-03
stateful_browser.py: add download_link method

moy:
This is my proposal for `download_link`, it superseeds #182 (I created a new PR to avoid force-pushing to a branch which is not mine).

### DIFF
--- a/docs/ChangeLog.rst
+++ b/docs/ChangeLog.rst
@@ -7,8 +7,9 @@ Version 1.0 (in development)
 
 Main changes:
 -------------
-* ``StatefulBrowser.follow_link`` and
-  ``StatefulBrowser.submit_selected()`` now sets the ``Referer:`` HTTP
+* ``StatefulBrowser.follow_link``,
+  ``StatefulBrowser.submit_selected()`` and the new
+  ``StatefulBrowser.download_link`` now sets the ``Referer:`` HTTP
   header to the page from which the link is followed.
 
 * ``Browser.submit`` and ``StatefulBrowser.submit_selected`` accept a larger
--- a/docs/ChangeLog.rst
+++ b/docs/ChangeLog.rst
@@ -16,6 +16,10 @@ Main changes:
   `requests.Session.request <http://docs.python-requests.org/en/master/api/#requests.Session.request>`__.
   [`#166 <https://github.com/MechanicalSoup/MechanicalSoup/pull/166>`__]
 
+* Added method ``StatefulBrowser.download_link``, which will download the
+  contents of a link to a file without changing the state of the browser.
+  [`#170 <https://github.com/MechanicalSoup/MechanicalSoup/issues/170>`__]
+
 Internal changes:
 -----------------
 * Private methods ``Browser._build_request`` and ``Browser._prepare_request``
--- a/mechanicalsoup/stateful_browser.py
+++ b/mechanicalsoup/stateful_browser.py
@@ -244,6 +244,36 @@ class StatefulBrowser(Browser):
         else:
             return links[0]
 
+    def _find_link_internal(self, link, args, kwargs):
+        """Wrapper around find_link that deals with convenience special-cases:
+
+        * If ``link`` has an *href*-attribute, then return it. If not,
+          consider it as a ``url_regex`` argument.
+
+        * If searching for the link fails and debug is active, launch
+          a browser.
+        """
+        if hasattr(link, 'attrs') and 'href' in link.attrs:
+            return link
+
+        # Check if "link" parameter should be treated as "url_regex"
+        # but reject obtaining it from both places.
+        if link and 'url_regex' in kwargs:
+            raise ValueError('link parameter cannot be treated as '
+                             'url_regex because url_regex is already '
+                             'present in keyword arguments')
+        else:
+            kwargs['url_regex'] = link
+
+        try:
+            return self.find_link(*args, **kwargs)
+        except LinkNotFoundError:
+            if self.get_debug():
+                print('find_link failed for', kwargs)
+                self.list_links()
+                self.launch_browser()
+            raise
+
     def follow_link(self, link=None, *args, **kwargs):
         """Follow a link.
 
@@ -260,16 +290,7 @@ class StatefulBrowser(Browser):
 
         :return: Forwarded from :func:`open_relative`.
         """
-        if not hasattr(link, 'attrs') or 'href' not in link.attrs:
-            try:
-                link = self.find_link(link, *args, **kwargs)
-            except LinkNotFoundError:
-                if self.get_debug():
-                    print('follow_link failed for', link)
-                    self.list_links()
-                    self.launch_browser()
-                raise
-
+        link = self._find_link_internal(link, args, kwargs)
         referer = self.get_url()
         request_kwargs = dict()
         if referer is not None:
--- a/mechanicalsoup/stateful_browser.py
+++ b/mechanicalsoup/stateful_browser.py
@@ -316,7 +316,11 @@ class StatefulBrowser(Browser):
         """
         link = self._find_link_internal(link, args, kwargs)
         url = self.absolute_url(link['href'])
-        response = self.session.get(url)
+        referer = self.get_url()
+        headers = dict()
+        if referer is not None:
+            headers["Referer"] = referer
+        response = self.session.get(url, headers=headers)
         if self.raise_on_404 and response.status_code == 404:
             raise LinkNotFoundError()
 
--- a/mechanicalsoup/stateful_browser.py
+++ b/mechanicalsoup/stateful_browser.py
@@ -298,6 +298,35 @@ class StatefulBrowser(Browser):
 
         return self.open_relative(link['href'], **request_kwargs)
 
+    def download_link(self, link=None, file=None, *args, **kwargs):
+        """Downloads the contents of a link to a file. This function behaves
+        similarly to :func:`follow_link`, but the browser state will
+        not change when calling this function.
+
+        :param file: Filesystem path where the page contents will be
+            downloaded. If the file already exists, it will be overwritten.
+
+        Other arguments are the same as :func:`follow_link` (``link``
+        can either be a bs4.element.Tag or a URL regex, other
+        arguments are forwarded to :func:`find_link`).
+
+        :return: `requests.Response
+            <http://docs.python-requests.org/en/master/api/#requests.Response>`__
+            object.
+        """
+        link = self._find_link_internal(link, args, kwargs)
+        url = self.absolute_url(link['href'])
+        response = self.session.get(url)
+        if self.raise_on_404 and response.status_code == 404:
+            raise LinkNotFoundError()
+
+        # Save the response content to file
+        if file is not None:
+            with open(file, 'wb') as f:
+                f.write(response.content)
+
+        return response
+
     def launch_browser(self, soup=None):
         """Launch a browser to display a page, for debugging purposes.
 
--- a/tests/test_stateful_browser.py
+++ b/tests/test_stateful_browser.py
@@ -402,5 +402,26 @@ def test_referer_submit_headers(httpbin):
     assert headers['X-Test-Header'] == 'x-test-value'
 
 
+def test_link_arg_text(httpbin):
+    browser = mechanicalsoup.StatefulBrowser()
+    browser.open_fake_page('<a href="/get">Link</a>', httpbin.url)
+    browser.follow_link(link_text='Link')
+    assert browser.get_url() == httpbin + '/get'
+
+
+def test_link_arg_regex(httpbin):
+    browser = mechanicalsoup.StatefulBrowser()
+    browser.open_fake_page('<a href="/get">Link</a>', httpbin.url)
+    browser.follow_link(url_regex='.*')
+    assert browser.get_url() == httpbin + '/get'
+
+
+def test_link_arg_multiregex(httpbin):
+    browser = mechanicalsoup.StatefulBrowser()
+    browser.open_fake_page('<a href="/get">Link</a>', httpbin.url)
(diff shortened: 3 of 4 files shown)


tests/test_stateful_browser.py
```diff
@@ -351,5 +353,65 @@ def test_select_form_nr():
...
+    assert browser.get_current_page() == current_page
+
+    # Check that the file was not downloaded
+    assert not os.path.exists(tmpfile)
+
+
+def test_download_link_referer(httpbin):
+    """Test downloading the contents of a link to file."""
+    browser = mechanicalsoup.StatefulBrowser()
+    ref = httpbin.url + "/my-referer"
```
hemberger: Very minor, but the `httpbin` object implements an `operator+` for strings, you can just use `httpbin + "my-referer"` to achieve the same thing. Feel free to ignore. :)

tests/test_stateful_browser.py
```diff
@@ -351,5 +353,65 @@ def test_select_form_nr():
...
+    browser.open_fake_page('<a href="/headers">Link</a>',
+                           url=ref)
+    tmpfile = tempfile.NamedTemporaryFile()
+    current_url = browser.get_url()
+    current_page = browser.get_current_page()
+    browser.download_link(tmpfile.name, link_text='Link')
```
hemberger: It wasn't in my original implementation, but thinking about it some more, it would be nice for `download_link` to return the response. I don't think there are any downsides to doing this.

If we do, then we can simplify this test. No need to read the downloaded file; instead, we can simply check the headers in the returned response.

mechanicalsoup/stateful_browser.py
```diff
@@ -237,7 +237,7 @@ def find_link(self, *args, **kwargs):
         else:
             return links[0]
 
-    def follow_link(self, link=None, *args, **kwargs):
+    def follow_link(self, link=None, return_url_only=False, *args, **kwargs):
```
hemberger: I mentioned it in the other PR, and I still think it would be confusing for `follow_link` to have an option that makes it not follow a link and instead do something that seems more like a behavior of `find_link`.

I would be strongly in favor of removing this and changing the implementation of `download_link` (see my suggestion below), unless you can convince me that this new option is a good idea!

mechanicalsoup/stateful_browser.py
```diff
@@ -262,7 +266,34 @@ def follow_link(self, link=None, *args, **kwargs):
...
+        """Downloads the contents of a link to a file. The browser state
+        will not change when calling this function.
+
+        :param filename: Filesystem path where the page contents will be
+            downloaded.
+        :param \*args, \*\*kwargs: Arguments forwarded to :func:`follow_link`.
+            These specify a link to the page whose contents you want to
+            download.
+        """
+        url = self.follow_link(return_url_only=True, *args, **kwargs)
```
hemberger: As per my comment above, I would rather see this as something like the following to avoid adding an option to `follow_link` that feels unnatural to me:

```python
    url = self.find_link(*args, **kwargs)['href']
```
Doing this means `download_link` can't take a bs4 tag as the link, which I'm fine with; however, if we demand it, then I think we can just reuse (or factor out) the "link-or-tag" input logic from `follow_link`, e.g.:

```python
def download_link(self, filename, link=None, *args, **kwargs):
    if not hasattr(link, 'attrs') or 'href' not in link.attrs:
        link = self.find_link(link, *args, **kwargs)
    url = link['href']

Comment (moy) [reply mechanicalsoup/stateful_browser.py]: You'll also need to deal with failure cases (follow_link has a special case to launch a browser when the debug mode is active, which is really convenient).

My first version (which I didn't push) factored this logic in a `find_link_or_fail` method. But in the end, the code was split into too many trivial methods to my taste.

File: mechanicalsoup/stateful_browser.py (hunk 3 above)
Comment (moy) [reply mechanicalsoup/stateful_browser.py]: I agreed with your point about not having a `download_to` parameter, because this would be a feature that the end user would really want to use and hiding it within the list of parameters of `follow_link` made it too hard to discover.

I disagree with this case: it's just a "dry-run" parameter, that asks the function to evaluate its functional part but to avoid doing side effects. I preferred calling it `return_url_only` to `dry_run` because the later doesn't say what the function is returning.

It's essentially a private feature, the user doesn't need to know about it (I still documented it in the public API in case someone wanted to use it, but we could also strip the doc and call it `_return_url_only`).

File: tests/test_stateful_browser.py (hunk 1 above)
Comment (moy) [reply tests/test_stateful_browser.py]: Done.

File: mechanicalsoup/stateful_browser.py (hunk 3 above)
Comment (moy) [reply mechanicalsoup/stateful_browser.py]: I agreed with your point about not having the download functionality hidden in the list of arguments of `follow_link`, but this one is different. It's not an option really meant for the end user (I still documented it just in case, but we could as well strip the doc and call it `_return_url_only` to make it clear that it's not meant for the end user).

It's basically a "dry run" option: it tells the function do evaluate its functional part but to avoid doing any side effects. I preferred calling it `return_url_only` to `dry_run` because the later doesn't tell what the function would return.

File: tests/test_stateful_browser.py (hunk 2 above)
Comment (moy) [reply tests/test_stateful_browser.py]: I agree that returning the response is nice, but we must still read the file to check that the writing happened correctly. Actually, we don't in this version, but we will in my next push.

Comment (moy) [reply tests/test_stateful_browser.py]: BTW, the existing code was creating the file before the download, and then asserting that the file existed, which was the case even without the download. I'm adding one more test to check that downloading to a non-existing file works, and that downloading to an existing file overwrites the file (we may want to provide other options than overwriting, but I want to test it to see if we change the behavior later, on purpose or by accident).

**Comment by moy** [comment]
OK, so I just did a new push with your suggestions, except the `return_url_only` part for which I disagree (see my comments above).

File: tests/test_stateful_browser.py (hunk 2 above)
Comment (hemberger) [reply tests/test_stateful_browser.py]: Great! Thanks for those improvements to the tests.

File: mechanicalsoup/stateful_browser.py (hunk 3 above)
Comment (hemberger) [reply mechanicalsoup/stateful_browser.py]: I have two primary concerns about such a parameter.

First, one might ask why we have a `dry_run`-like parameter for `follow_link`, but nothing else? Along the same lines, if I saw a `verbose` or `debug` argument for one function, I might naturally expect it to exist for most other functions, because it represents an infrastructural design choice or philosophy. But in this case, it is just one ad-hoc parameter.

Second, let us consider the possible scenarios for such a parameter:
1. The input `link` is a bs4 tag, in which case `follow_link` becomes an opaque wrapper for `link['href']`. 
2. The input `link` is not a bs4 tag, in which case `follow_link` becomes a wrapper for `find_link` that returns `link['href']` instead of `link`.

Neither of these scenarios seem like appropriate behavior for a method called `follow_link`, especially when the same functionality can be achieved using existing methods. I don't like the prospect of distorting the public API to fulfill an internal need. But even if the parameter were private, it just doesn't seem a logical thing to do.

I don't know, am I being too difficult about something that ultimately doesn't matter?

**Comment by hemberger** [comment]
I can't really tell, but would the `stream=True` parameter avoid making extra copies of `response.content` by allowing us to stream directly to the file, or would it just delay it? Does it matter either way?

**Comment by moy** [comment]
Streaming the result directly to the file would certainly be better (loading everything in the RAM to write it to a file is certainly a brute-force approach...). Ideally, we could allow the browser to continue browsing while the download streams to the file.

I won't do this in this PR, but it can be added later if needed.

File: mechanicalsoup/stateful_browser.py (hunk 3 above)
Comment (moy) [reply mechanicalsoup/stateful_browser.py]: Actually, there's another rather simple option: move most of the logic of `follow_link` (launching a browser in debug mode and special-casing elements with an href attribute) to `find_link`. Then `find_link` becomes a straightforward wrapper, and `download_link` remains simple too. I'll push something along these lines shortly.

**Comment by moy** [comment]
Just pushed a version that should address all your remarks (+ rebased on top of the Referer PR).

File: mechanicalsoup/stateful_browser.py [hunk 5]
```diff
@@ -230,16 +230,32 @@ def links(self, url_regex=None, link_text=None, *args, **kwargs):
                          if a.text == link_text]
         return all_links
 
-    def find_link(self, *args, **kwargs):
+    def find_link(self, link=None, *args, **kwargs):
```
hemberger: Won't this break backwards compatibility for anyone who has specified `*args`? Unfortunately this whole situation is muddled by the `*args` situation, which doesn't seem to have a satisfying resolution. :/

Actually, in this one case, I think we can at least not make the situation worse by keeping
```python
if not hasattr(link, 'attrs') or 'href' not in link.attrs:
    link = self.find_link(*args, **kwargs)
```
in both `follow_link` and `download_link` instead of
```python
link = self.find_link(link, *args, **kwargs)
```
It's only 1 line of code duplicated, but I think avoids a lot of complexity in `find_link`.

mechanicalsoup/stateful_browser.py
```diff
@@ -230,16 +230,32 @@ def links(self, url_regex=None, link_text=None, *args, **kwargs):
...
+
         If no link is found, raise :class:`LinkNotFoundError`.
         """
+        if hasattr(link, 'attrs') and 'href' in link.attrs:
+            return link
+        if link is not None:
+            args = (link,) + args
         links = self.links(*args, **kwargs)
         if len(links) == 0:
+            if self.get_debug():
```
hemberger: No change requested here. Just wanted to say that I like the idea of moving this debug conditional from `follow_link` to `find_link`.

mechanicalsoup/stateful_browser.py
```diff
@@ -260,23 +276,49 @@ def follow_link(self, link=None, *args, **kwargs):
...
+        similarly to :func:`follow_link`, but the browser state will
+        not change when calling this function.
+
+        :param filename: Filesystem path where the page contents will be
+            downloaded. If the file already exists, it will be overwritten.
+        :param link: If this parameter has an *href*-attribute,
+            download the page pointed-to by this attribute. Otherwise,
+            consider this as the first argument forwarded
+            to :func:`find_link`.
+        :param \*args&\*\*kwargs: Arguments forwarded
```
hemberger: The description of these arguments is a bit obsolete with the addition of the `link` param. Perhaps we can just leave it as "Additional arguments forwarded to find_link"?

docs/ChangeLog.rst
```diff
@@ -16,6 +16,10 @@ Main changes:
   `requests.Session.request <http://docs.python-requests.org/en/master/api/#requests.Session.request>`__.
   [`#166 <https://github.com/MechanicalSoup/MechanicalSoup/pull/166>`__]
 
+* Added method ``StatefulBrowser.download_link``, which will download the
```
hemberger: Should we add `download_link` to the list of methods in the Referer bullet point?

mechanicalsoup/stateful_browser.py
hemberger: (And, if it wasn't clear, we would not need the `link` param in `find_link` in this case.)

moy: > Won't this break backwards compatibility for anyone who has specified *args?

Not really. `*args` is for positional arguments, so what I did is to special-case the first positional argument. A call to `find_link(foo, bar, boz)` was earlier setting `args=(foo, bar, boz)` and now `link=foo, args=(bar, boz)`. But since I later forward `(link,) + args`, the result is the same.

I can think of one case where this was not backward-compatible: `find_link(None, foo)`, i.e. pass `None` as a positional argument and expect it to be forwarded as `url_regex` (so that `foo` becomes the `link_text` argument). I'll fix it in my next push.

hemberger: I really think any change to the `find_link` interface is overcomplicating this...

moy: I actually think that any difference between `find_link` and `list_link`'s list of argument is a source of problem. Changing the first to match the second will fix #178 as a side effect.

moy:
OK, I finally went a different way, as I could not find a better way to get this `download_link` and fix #177.

It looks like your suggestion with a bit of duplicated code, except that the fix to #177 adds much more code to potentially duplicate in `follow_link`, so I created a helper function. This function ends up looking a lot like what I tried to do in `find_link`, but it's an internal one to deal with convenience special-cases, so it doesn't need to change the existing `find_link`.

mechanicalsoup/stateful_browser.py
```diff
@@ -298,6 +298,34 @@ def follow_link(self, link=None, *args, **kwargs):
...
+            <http://docs.python-requests.org/en/master/api/#requests.Response>`__
+            object.
+        """
+        link = self._find_link_internal(link, args, kwargs)
+        url = self.absolute_url(link['href'])
+        response = self.session.get(url)
+        if self.raise_on_404 and response.status_code == 404:
+            raise LinkNotFoundError()
+
+        # Save the response content to file
```
amotl: Dear @moy, thanks for your work on this!

May i humbly suggest to make writing to the filesystem optional? My proposal would be to change the method signature to:
`def download_link(self, link=None, filename=None, *args, **kwargs)`

... and then conditionally run the code like:
```
if filename is not None:
    with open(filename, 'wb') as f:
        f.write(response.content)
```

----

A further improvement could be to accept an already opened file handle as value for the `filename` argument like
```
if filename is not None:
    if isinstance(filename, six.string_types):
        with open(filename, 'wb') as f:
            f.write(response.content)
    elif hasattr(filename, 'write'):
        filename.write(response.content)
```

When doing this, one would be able to also pass an in-memory string buffer like an instance of `StringIO` to the `download_link` method.

amotl: Thanks!

hemberger:
This looks good to go for me!

amotl:
Thanks Matthieu. Works like a charm!

Issue #175, opened 2017-12-07
Shorter getter names


hemberger:
I've always been a little bothered having to write `browser.get_current_page()` and `browser.get_current_form()`. They're quite lengthy.

With `browser.get_current_page()`, I imagine that `get_page` would suffice (and perhaps there are even shorter aliases we could entertain). There really is not ever another page, so "current" is a superfluous adjective.

With `browser.get_current_form()`, if we're being verbose, then `get_selected_form` seems like it would be more correct and also less confusing to users. However, `get_selected_form` is quite lengthy as well, and I'm not convinced that `get_form` is suitable due to the (rightful) existence of `select_form`.

moy:
Actually, we could/should even use Python's `@property` and avoid getters right away.

moy:
Indeed, I messed up with a push (which I cancelled immediately but it already applied the "fixes" directive).

The current pr is #294.

PR #185, opened 2018-01-04
Add more succinct state access options

hemberger:
This PR combines a few changes, mostly proposed in #175.

* Deprecate longer/misleading getter names.
  `get_current_page` -> `get_page` (since there's no other page than the current one)
  `get_current_form` -> `get_selected_form` (to avoid confusion about what form this returns)

* Add properties for convenient state access
  `page` == `get_page()`
  `url` == `get_url()`
  `form` = `get_selected_form()` (rather than `selected_form` for maximum convenience)

* To combat any possible misinterpretation of the `form` property (and because I think it makes the workflow more transparent), `get_selected_form` will raise an exception if no form has been selected. I think that "Why is `browser.form` None?" is a good question to preempt.

There are a few things to decide if we want to move forward with this PR:
1. Since we are using properties, do we still want to change the getter names? I would say yes, but I'm open to alternatives.
2. Do we want complete backwards compatibility? If so, then perhaps we can consider keeping (but still deprecating) `get_current_form` with the exception-free behavior instead of aliasing it to `get_selected_form`. However, I think the exception version is safer and a fairly transparent backwards incompatibility, and I would prefer transitioning to the new version outright.
3. Since the properties are so much more convenient, should we advertise them in the docs in favor of the traditional getters?

### DIFF
--- a/docs/ChangeLog.rst
+++ b/docs/ChangeLog.rst
@@ -12,6 +12,16 @@ Main changes:
   ``StatefulBrowser.download_link`` now sets the ``Referer:`` HTTP
   header to the page from which the link is followed.
 
+* ``StatefulBrowser`` methods ``get_current_page`` and ``get_current_form``
+  have been deprecated in favor of ``get_page`` and ``get_selected_form``
+  respectively. However, you can now access the browser state more conveniently
+  with attributes ``page`` (calls ``get_current_page()``), ``form`` (calls
+  ``get_selected_form()``), and ``url`` (calls ``get_url()``).
+  [`#175 <https://github.com/MechanicalSoup/MechanicalSoup/issues/175`__]
+
+* ``StatefulBrowser.get_selected_form`` will now raise an ``AttributeError``
+  istead of returning ``None`` if no form has been selected yet.
+
 * ``Browser.submit`` and ``StatefulBrowser.submit_selected`` accept a larger
   number of keyword arguments. Arguments are forwarded to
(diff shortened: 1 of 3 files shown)


moy:
>     Since we are using properties, do we still want to change the getter names? I would say yes, but I'm open to alternatives.

I don't think we need the getters. The point of properties is to avoid them or make them transparent.

So, I'd do:

* Add the properties. This should look like
```
    @property
    def form(self):
        return self.__state.form
```
* Keep the existing getters, but deprecate them
* No need to create new getters with shorter names

>     Do we want complete backwards compatibility? If so, then perhaps we can consider keeping (but still deprecating) get_current_form with the exception-free behavior instead of aliasing it to get_selected_form. However, I think the exception version is safer and a fairly transparent backwards incompatibility, and I would prefer transitioning to the new version outright.

I can imagine someone writing:

    if browser.get_current_form() == None:
        browser.select_form(...)

That's reasonably unlikely, but it probably makes sense to keep the existing getter as-is.

>     Since the properties are so much more convenient, should we advertise them in the docs in favor of the traditional getters?

To me, yes.

hemberger:
Okay, thanks for the feedback! I'll get to those changes soon.

I've always liked getters, because they provide a natural way to group properties at a functional level (and if I want to see them all, I just type `.get<Tab>` in ipython), but I suppose you're right, as usual, that `@property` is the way to go. :)

hemberger:
This PR is superseded by #294.

PR #186, opened 2018-01-04
Add PatZilla to tools using MechanicalSoup

hemberger:
@amotl: do you mind if we add your tool to our "using MechanicalSoup" list?

### DIFF
--- a/docs/external-resources.rst
+++ b/docs/external-resources.rst
@@ -34,4 +34,6 @@ look at their source code for real-life examples.
 * `Chamilo Tools <https://gitlab.com/chamilotools/chamilotools>`__
 * `gmusicapi <https://github.com/simon-weber/gmusicapi>`__: an unofficial API
   for Google Play Music
+* `PatZilla <https://github.com/ip-tools/ip-navigator>`__: Patent information
+  research for humans
 * *TODO: Add your favorite tool here ...*


amotl:
Please go ahead, appreciate it. Thanks again for your responsiveness and excellent work!

amotl:
Dear @hemberger and @moy,

i see you are still working on the upcoming 1.0.0-dev release, right? Maybe you could upload another interim version like 0.9.0.post5 to PyPI? That would be cool as we are experiencing problems with the `dependency_links` thing when installing _PatZilla_ with pip directly from PyPI.

Just doing `python setup.py develop` works for us but this is only suitable for development. There are the usual shenanigans regarding setuptools/pip compatibility around on the internet with hints about the parameters like `--process-dependency-links --allow-all-external --allow-unverified=`, which we tried to apply. However, we haven't had any success with this.

Thanks again for your efforts!

WIth kind regards,
Andreas.

hemberger:
I would be happy to do a 0.9.1 or 0.10.0 release (or post5, if we're only cherry-picking a small change). I'll coordinate with @moy and report back.

moy:
My vote would be:

1) Release 0.10.0 from master
2) wait a bit
3) if no one complains, release 1.0

amotl:
Hi there,

we just wanted to let you know that we just switched to the 0.10.0 release (https://github.com/ip-tools/ip-navigator/commit/b12b2c15) and everything works as expected.

Thanks again for your efforts,
Andreas.

Issue #187, opened 2018-01-05
Page refresh functionality


tbolender:
It would be nice to have a refresh functionality which reloads the current page with the same request method and parameters. This is handy when waiting/looking for a certain change on a website to appear.

This would furthermore complete the goal of implementing an actual browser as proposed in #127.

In my opinion saving the last request sent and resending it should be sufficient to implement this functionality. Unfortunately, if I understood correctly, there seems to be no function which every request passes?

hemberger:
Well, all requests are fundamentally a `Requests.PreparedRequest` object, which you can access from `response.request`. We could potentially store the `response` (or the `response.request`) in the browser's state, and then I believe `StatefulBrowser.refresh` would simply be:
```python
def refresh(self):
    self.session.send(self._state.response.request)
```
(with some extra infrastructure to check the result, store the new state, etc.)

Unless there's something I'm missing here, this seems like a pretty easy (and very useful) enhancement.

tbolender:
That sounds nice! I did not know that you could retrieve the request from a response. Thereby also usages of the thin wrappers `get()` and `post()` could easily be handled (this caused some worries when looking into this myself).

PR #188, opened 2018-01-05
Page refresh

tbolender:
I started to implement the page refresh functionality as described in #187. Basically, I added the request to the `_BrowserState` and using it to resend the last request.

Before implementing any tests (so coverage is currently down), I would like to get some feedback whether I missing something or anything is not correctly implemented.

### DIFF
--- a/mechanicalsoup/stateful_browser.py
+++ b/mechanicalsoup/stateful_browser.py
@@ -10,10 +10,11 @@ import bs4
 
 
 class _BrowserState:
-    def __init__(self, page=None, url=None, form=None):
+    def __init__(self, page=None, url=None, form=None, request=None):
         self.page = page
         self.url = url
         self.form = form
+        self.request = request
 
 
 class StatefulBrowser(Browser):
@@ -130,7 +131,8 @@ class StatefulBrowser(Browser):
             print(url)
 
         resp = self.get(url, *args, **kwargs)
-        self.__state = _BrowserState(page=resp.soup, url=resp.url)
+        self.__state = _BrowserState(page=resp.soup, url=resp.url,
+                                     request=resp.request)
         return resp
 
     def open_fake_page(self, page_text, url=None, soup_config=None):
@@ -200,7 +202,8 @@ class StatefulBrowser(Browser):
 
         resp = self.submit(self.__state.form, url=self.__state.url,
                            *args, **kwargs)
-        self.__state = _BrowserState(page=resp.soup, url=resp.url)
+        self.__state = _BrowserState(page=resp.soup, url=resp.url,
+                                     request=resp.request)
         return resp
 
     def list_links(self, *args, **kwargs):
--- a/mechanicalsoup/stateful_browser.py
+++ b/mechanicalsoup/stateful_browser.py
@@ -153,6 +153,27 @@ class StatefulBrowser(Browser):
         """
         return self.open(self.absolute_url(url), *args, **kwargs)
 
+    def refresh(self):
+        """Reload the current page with the same request as originally done.
+        Any change (`select_form`, or any value filled-in in the form) made to
+        the current page before refresh is discarded.
+
+        :raise ValueError: Raised if no refreshable page is loaded, e.g., when
+            using the shallow ``Browser`` wrapper functions.
+
+        :return: Response of the request."""
+        old_request = self.__state.request
+        if old_request is None:
+            raise ValueError('The current page is not refreshable. Either no '
+                             'page is opened or low-level browser methods '
+                             'were used to do so')
+
+        resp = self.session.send(old_request)
+        Browser.add_soup(resp, self.soup_config)
+        self.__state = _BrowserState(page=resp.soup, url=resp.url,
+                                     request=resp.request)
+        return resp
+
     def select_form(self, selector="form", nr=0):
         """Select a form in the current page.
 
--- a/tests/test_stateful_browser.py
+++ b/tests/test_stateful_browser.py
@@ -6,7 +6,7 @@ import mechanicalsoup
 import sys
 import re
 from bs4 import BeautifulSoup
-from utils import setup_mock_browser
+from utils import setup_mock_browser, prepare_mock_browser, mock_get
 import pytest
 import webbrowser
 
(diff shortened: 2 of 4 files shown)


hemberger:
That's pretty much exactly what I envisioned. I hope it works as expected!

I would be happy with this implementation, but I'd be curious to hear @moy's opinion, since he is often perceptive to things that I might overlook.

mechanicalsoup/stateful_browser.py
```diff
@@ -152,6 +152,24 @@ def open_relative(self, url, *args, **kwargs):
         """
         return self.open(self.absolute_url(url), *args, **kwargs)
 
+    def refresh(self):
+        """Reloads the current page with the same request as originally done.
+        Works only if page was loaded using ``StatefulBrowser`` methods.
+        The form selection and values are not retained.
```
moy: I first mis-read this sentence as "the values in the form obtained to get to this page are discarded", which is incorrect.

Perhaps: "Any change (``select_form``, or any value filled-in in the form) made to the current page before refresh is discarded"?

mechanicalsoup/stateful_browser.py
```diff
@@ -152,6 +152,24 @@ def open_relative(self, url, *args, **kwargs):
         """
         return self.open(self.absolute_url(url), *args, **kwargs)
 
+    def refresh(self):
+        """Reloads the current page with the same request as originally done.
+        Works only if page was loaded using ``StatefulBrowser`` methods.
```
moy: I'd omit this "works only", since most methods of StatefulBrowser already behave like this.

mechanicalsoup/stateful_browser.py
```diff
@@ -152,6 +152,24 @@ def open_relative(self, url, *args, **kwargs):
         """
         return self.open(self.absolute_url(url), *args, **kwargs)
 
+    def refresh(self):
+        """Reloads the current page with the same request as originally done.
```
moy: We usually write docstring with imperative tone, hence s/Reloads/Reload/

mechanicalsoup/stateful_browser.py
```diff
@@ -152,6 +152,24 @@ def open_relative(self, url, *args, **kwargs):
...
+        Works only if page was loaded using ``StatefulBrowser`` methods.
+        The form selection and values are not retained.
+
+        :raise ValueError: Raised if no refreshable page is loaded, e.g., when using
+            the shallow ``Browser`` wrapper functions.
+
+        :return: Response of the request."""
+        old_request = self.__state.request
+        if old_request is None:
+            raise ValueError('The current page is not refreshable')
```
moy: A hint on why this fails would be welcome. But we can also merge this as-is and improve the message later if needed.

mechanicalsoup/stateful_browser.py
hemberger: I think he was referring to the fact that it won't reload the page from a `StatefulBrowser.get`, even though called on a `StatefulBrowser` object, since it's using a `Browser` (stateless) method. I'm not sure how to best phrase this.

I almost want to get rid of the `Browser` class and have everything be "stateful" to avoid this kind of ambiguity, but... that's a much longer discussion for a much later time. ;)

tbolender: @hemberger is right about my intention. Until I started to created the PR I did not know that there was a different behavior when opening pages via `StatefulBrowser` or `Browser`. Maybe I just over-read this somewhere.

And yes, I would also support to get rid of `Browser` :D

tbolender:
Tests are self-evident, I just did not want to put more effort in if somebody disagrees with the basic implementation ;) I will implement them as soon as I have time.

Rebasing will be done as well :)

I thought about storing the full response, but this would have reduced `_BrowserState` to a response and form. From the perspective of a clean history, I would apply such a change in a different PR, but could also do it within this one.

mechanicalsoup/stateful_browser.py
moy: > I think he was referring to the fact that it won't reload the page from a StatefulBrowser.get

Yes, but so does following a relative link or submitting a form with a relative action. It's not specific to `reload`.

> I almost want to get rid of the Browser class and have everything be "stateful" to avoid this

Let's start a separate discussion: #190

mechanicalsoup/stateful_browser.py
tbolender: I am not sure about this one. I extended the description, but in the end we have something similar to the not-working hint above.

mechanicalsoup/stateful_browser.py
tbolender: Since there is currently no other hint of a distinction between `StatefulBrowser` and `Browser` methods I decided to omit the whole sentence.

tbolender:
I addressed the documentation issues and created the some tests.

If I understood correctly, you mainly rely on `open_fake_page()` for testing. Unfortunately, opening a page this way this does not cause the creation of a request method which makes testing infeasible. I noticed though that there exists a `setup_mock_browser()` which prepares some actual mocked requests. Since I did not want to change much of the existing code I wrote my own helper function. I hope this is fine with you.

tbolender:
You are welcome! I added entries to the changelog as well. Just tell me if you prefer a different wording.

tests/test_stateful_browser.py
```diff
@@ -532,5 +532,73 @@ def test_download_link_referer(httpbin):
...
+
+    browser, adapter = prepare_mock_browser()
+    mock_get(adapter, url=url, reply=str(initial_page))
+    browser.open(url)
+    mock_get(adapter, url=url, reply=str(reload_page))
+
+    browser.refresh()
+
+    assert browser.get_url() == url
+    assert browser.get_current_page() == reload_page
```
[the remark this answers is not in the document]
tbolender: This is indeed the behavior of Firefox 57.0.4, tested with [this gist](https://gist.github.com/phrawzty/62540f146ee5e74ea1ab). I added respective checks.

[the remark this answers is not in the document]
moy: Cool! All good now, let's merge this.

PR #195, opened 2018-02-09
Handles html entities that don't say they are html

dannypurcell:
One of the web servers I tried MechanicalSoup against leaves out the `Content-Type` on some HTML containing responses.
According to https://tools.ietf.org/html/rfc2616#section-7.2.1 
this is actually valid behavior and we are expected to guess if we so choose.
This changes `add_soup(response, soup_config)` to take a swag in that case.

### DIFF
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -52,15 +52,20 @@ class Browser(object):
                 self.session.mount(adaptee, adapter)
 
         self.soup_config = soup_config or dict()
-    
+
+    @staticmethod
     def __looks_like_html(response):
-        text = response.text.lstrip()
-        return text.startswith('<html') or text.startswith('<!DOCTYPE')
+        """Guesses entity type when Content-Type header is missing.
+        Since Content-Type is not strictly required, some servers leave it out.
+        """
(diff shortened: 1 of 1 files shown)


dannypurcell:
@hemberger fixed, and Travis is happy now.
Also moved the guesswork to a separate function so we can avoid `lstrip` when not needed

hemberger:
I'm going to give @moy a few days to review this if he wants, and then I'll merge it into master. Thanks again for your contribution!

mechanicalsoup/browser.py
```diff
@@ -53,10 +53,19 @@ def __init__(self, session=None, soup_config={'features': 'lxml'},
 
         self.soup_config = soup_config or dict()
 
+    @staticmethod
+    def __looks_like_html(response):
+        """Guesses entity type when Content-Type header is missing.
+        Since Content-Type is not strictly required, some servers leave it out.
+        """
+        text = response.text.lstrip()
+        return text.startswith('<html') or text.startswith('<!DOCTYPE')
```
moy: This could be case insensitive, since at least `<HTML` is possible too.

dannypurcell: fixed

PR #199, opened 2018-02-27
Remove `name` attribute from all unused buttons on form submit

blackwind:
I ran into a site with forms including buttons of `type` "button" with `name` attributes. Because `Form.choose_submit()` was only removing `name` from buttons of `type` "submit", the values for the "button" buttons were being erroneously sent on POST, thereby breaking my submission. This patch fixes the issue, even when a submit button isn't explicitly chosen.

Note that all buttons that aren't of `type` "button" or "reset" function as "submit" in all major browsers and should therefore be choosable.

### DIFF
--- a/mechanicalsoup/form.py
+++ b/mechanicalsoup/form.py
@@ -34,6 +34,7 @@ class Form(object):
 
     def __init__(self, form):
         self.form = form
+        self._submit_chosen = False
 
         # Aliases for backwards compatibility
         # (Included specifically in __init__ to suppress them in Sphinx docs)
@@ -286,7 +287,8 @@ class Form(object):
         """Selects the input (or button) element to use for form submission.
 
         :param submit: The bs4.element.Tag (or just its *name*-attribute) that
-            identifies the submit element to use.
+            identifies the submit element to use. If ``None``, will choose the
+            first valid submit element in the form, if one exists.
 
         To simulate a normal web browser, only one submit element must be
         sent. Therefore, this does not need to be called if there is only
@@ -303,33 +305,41 @@ class Form(object):
             form.choose_submit('form_name_attr')
             browser.submit_selected()
         """
+        # Since choose_submit is destructive, it doesn't make sense to call
+        # this method twice unless no submit is specified.
+        if self._submit_chosen:
+            if submit is None:
+                return
+            else:
+                raise Exception('Submit already chosen. Cannot change submit!')
 
-        submit = (
-            submit or
-            self.form.select_one('input[type="submit"]') or
-            self.form.select_one('button[type="submit"]')
-        )
+        # All buttons NOT of type (button,reset) are valid submits
+        inps = [i for i in self.form.select('input[type="submit"], button')
+                if i.get('type', '') not in ('button', 'reset')]
+
+        # If no submit specified, choose the first one
+        if submit is None and inps:
+            submit = inps[0]
 
         found = False
-        inps = self.form.select('input[type="submit"], button')
         for inp in inps:
-            if inp.get('type', "").lower() not in ("button", "reset") and (
-                inp == submit or (inp.has_attr('name') and
-                                  inp['name'] == submit)
-            ):
+            if inp == submit or (inp.has_attr('name') and
+                                 inp['name'] == submit):
                 if found:
                     raise LinkNotFoundError(
                         "Multiple submit elements match: {0}".format(submit)
                     )
                 found = True
-                continue
+            else:
+                # Delete any non-matching element's name so that it will be
+                # omitted from the submitted form data.
+                del inp['name']
 
-            del inp['name']
-
-        if not found:
+        if not found and submit is not None:
             raise LinkNotFoundError(
                 "Specified submit element not found: {0}".format(submit)
             )
+        self._submit_chosen = True
 
     def print_summary(self):
         """Print a summary of the form.
--- a/mechanicalsoup/form.py
+++ b/mechanicalsoup/form.py
@@ -304,11 +304,19 @@ class Form(object):
             browser.submit_selected()
         """
 
+        submit = (
+            submit or
+            self.form.select_one('input[type="submit"]') or
+            self.form.select_one('button[type="submit"]')
+        )
+
         found = False
-        inps = self.form.select('input[type="submit"], button[type="submit"]')
+        inps = self.form.select('input[type="submit"], button')
         for inp in inps:
-            if inp == submit or (inp.has_attr('name') and
-                                 inp['name'] == submit):
+            if inp.get('type', "").lower() not in ("button", "reset") and (
+                inp == submit or (inp.has_attr('name') and
+                                  inp['name'] == submit)
+            ):
                 if found:
                     raise LinkNotFoundError(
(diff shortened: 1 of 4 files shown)


blackwind:
Looks like the test suite doesn't like what I've done, but my solution works for me with the functions I use, and I don't have any further time to spend on my project that leverages this library at the moment. Hopefully someone can take my work here and run with it, as it's definitely a problem that needs solving.

hemberger:
Thanks for your contribution, @blackwind!

According to the HTML standard, it does indeed look like `type` defaults to "submit" if the `type` is unspecified or an invalid value is specified.

Unless someone else gets to it first, I'll write a few tests for these cases and see what changes are necessary. I think we probably want to avoid having `choose_submit` called by default, but other than that the patch seem fine to me (once we get the tests passing).

blackwind:
Did you ever get a chance to take a look at this, @hemberger?

hemberger:
So to reiterate: the [standard](https://www.w3.org/TR/2011/WD-html5-20110525/the-button-element.html#attr-button-type) says that if `type` is `button` or `reset`, it is treated like an input element; otherwise (if `type` is `submit`, unspecified, or not a valid value) it is treated as a submit element.

However, [w3schools](https://www.w3schools.com/tags/tag_button.asp) says that in practice this is not followed:
> Tip: Always specify the type attribute for a `<button>` element. Different browsers use different default types for the `<button>` element.

I'm going to assume that we'll just stick to the standard here, unless there are any objections.

blackwind:
In a perfect world, we'd stick with the standard, but the fact is, all major browsers (Chrome, Edge, Firefox, and IE) accept invalid values, so invalid values are being used in the wild -- including the website I'm scraping with this module. The module, therefore, needs to handle this use case.

hemberger:
I finally got around to adding the tests and fixing some of the issues that were causing this PR to fail existing tests. Sorry for the delay!

@blackwind Handling typeless buttons is part of the standard, so we're all good there! I was referring to:
> Different browsers use different default types for the `<button>` element.

But I don't know if that's true, and I don't know if that matters, since the standard says explicitly what to do (and this PR follows the standard). I'm happy with the result if you are.

@moy If you have a spare minute to look at my fixup commit message, there are a few changes that may or may not warrant comments in the changelog. Would appreciate your opinion here, if you're happy with the patch otherwise. Thanks!

mechanicalsoup/form.py
```diff
@@ -304,11 +304,19 @@ def choose_submit(self, submit):
             browser.submit_selected()
         """
 
+        submit = (
+            submit or
+            self.form.select_one('input[type="submit"]') or
+            self.form.select_one('button[type="submit"]')
+        )
+
```
moy: This is weird, or at least undocumented. The docstring says that `submit` must be a bs4.element.Tag or a string, so there should be no need to look for something else. If the goal is to let the function run when `submit` is not given, then 1) this should be documented, and 2) it would make sense to give a default value to the parameter, like `submit=None`. And then, since the behavior is otherwise to error out when there are multiple matches, I wouldn't do a "`or`" chain, but I'd check that there's one and only one match in `input[type="submit"],button[type="submit"]`. Picking an arbitrary button/submit is very fragile and shouldn't be encouraged.

mechanicalsoup/form.py
```diff
@@ -303,33 +305,41 @@ def choose_submit(self, submit):
             form.choose_submit('form_name_attr')
             browser.submit_selected()
         """
+        # Since choose_submit is destructive, it doesn't make sense to call
```
moy: Actually, I'd say this is a sign that `choose_submit` does not do the right thing. It's not documented to be destructive, and shouldn't be actually.

This `name` attribute stripping is in the wrong place IMHO. We should just get rid of this `del inp['name']` and ignore the corresponding elements in `_request`. Or did I miss anything?

mechanicalsoup/form.py
moy: Ah, the doc is here, but in the next commit. Either squash both commits together (probably makes sense), or re-split the commits properly.

hemberger:
I agree that the existing design is not ideal. We could in theory process the soup object in `Form.choose_submit` as we do now, and then instead of deleting any non-matching submits, just store the matching submit. Then, when `Browser.submit` is called, we could call a new function `Form.delete_unused_submits` or similar before extracting the soup object to pass to `Browser._request`. I did consider this, but I didn't want to add a reprocessing layer for a feature that has never been requested (`Form.choose_submit` has always been destructive in a way that prevents calling it twice -- it seemed helpful to at least document this with a clearer error message).

Unless there is a really compelling suggestion, my recommendation would be to evaluate the PR as-is and then consider refactoring `Form.choose_submit` as a separate task.

As for commit organization, I'd like to keep the original commit intact if you don't mind too much. Were they only my commits, I'd certainly agree to squash, but I don't think it muddies up the history too much here, and it's nice to recognize blackwind's contribution. I usually just look at the "Files Changed" tab to see the overall effect of the PR, and then only delve into the individual commits if something isn't clear there. Would that be okay?

mechanicalsoup/form.py
```diff
@@ -303,9 +305,23 @@ def choose_submit(self, submit):
...
+            else:
+                raise Exception('Submit already chosen. Cannot change submit!')
+
+        # All buttons NOT of type (button,reset) are valid submits
+        inps = [i for i in self.form.select('input[type="submit"], button')
+                if i.get('type', '') not in ('button', 'reset')]
+
+        # If no submit specified, choose the first one
+        if submit is None and inps:
+            submit = inps[0]
```
moy: That's better than the `or`-chain I saw in the first commit, but I think this should raise when `len(inps) > 1`.

hemberger:
@moy I'm happy to merge this PR as-is, but I do want to address your comments in case there is room for improvement. A few points:
* I don't think there are any significant edge-cases in this code than there were before, but previously they were undocumented and uncaught (you would just get some indirect error, which would require you to go into the source code to figure out what it was doing). With this PR, all these edge cases are now tested too, at the expense of a few more lines of code.
* The name attribute stripping occurs in `Form.choose_submit` instead of `Browser._request` because there is no interface at the `Browser`-level to specify which submit to choose. If we want to change the `Browser` interface, I agree it would allow us to simplify the complicated (and, I agree, not ideal) layout of `Form.choose_submit`, but I think it would be a non-trivial backwards incompatibility.

So I see three paths forward:
1. Accept the flaws in the current code and merge this PR.
2. Merge this PR, but plan to modify `Form` and `Browser` to simplify choosing submits at some unspecified time in the future.
3. Decide on a plan to modify `Form` and `Browser` now and abandon this PR.

My preferences are (in order) 2, 1, 3. What do you think?

moy:
Agreed for 2, 1, 3.

moy:
Unless someone objects soon, I'll merge this.

moy:
OK, let's merge. As much as possible, we should improve the code later.

hemberger:
It's odd that this passed tests here, but now that it's merged it doesn't anymore...

moy:
Indeed. Do you want to debug this or shall I?

hemberger:
I've got a simple fix, which I'll submit as PR now.

PR #208, opened 2018-06-03
faq.rst: address common Stack Overflow question

hemberger:
This answers the question of what to do when form submission
fails due to the website using JavaScript.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -68,6 +68,8 @@ variable. :class:`mechanicalsoup.StatefulBrowser` is essentially a
 superset of :class:`mechanicalsoup.Browser`, it's the one you should
 use unless you have a good reason to do otherwise.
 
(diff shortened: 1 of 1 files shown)


moy:
Looks good to me. I'm wondering whether we should issue a warning when we detect no action= field and JavaScript to be executed on submisson.

There's a Travis-CI failure which deserves investigation but seems unrelated.

hemberger:
I've fixed the tests. The failures were caused by network issues with httpbin.org. Since the Travis pypy environments now work with `pytest-httpbin`, I've merged that in and rebased this PR.

As for a warning when no `action` attribute is detected, that could get annoying in cases where there is intentionally no `action` (since the HTML standard allows it to be omitted if the action is the current page). I'm not sure how to detect JavaScript interference without false positives, but if you have any ideas that certainly sounds useful.

PR #216, opened 2018-06-27
Make sure all tests can be run offline

### LINKED ISSUE #215: httpbin tests require network access
Hi,
It seems that tests require network access. You modified pytest-httpbin to make it connect to the real website instead of the local version, why is that?
The Gentoo sandbox doesn't allow network access so this causes the tests to fail:
```
E           requests.exceptions.ConnectionError: HTTPConnectionPool(host='httpbin.org', port=80): Max retries exceeded with url: /forms/post (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f8ca79b8ac8>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution',))

```

Here is the full [build.log](https://github.com/MechanicalSoup/MechanicalSoup/files/2141788/build.log).

hemberger:
Fixes #215.

There were three tests that were still accessing http://httpbin.org.
By switching these to use the `httpbin` pytest fixture, we no longer
require any network access.

### DIFF
--- a/tests/test_browser.py
+++ b/tests/test_browser.py
@@ -37,33 +37,33 @@ def test_submit_online(httpbin):
     assert 'MechanicalSoup' in json["headers"]["User-Agent"]
 
 
-form_html = """
-<form method="post" action="http://httpbin.org/post">
-<input name="customer" value="Philip J. Fry"/>
-<input name="telephone" value="555"/>
-<textarea name="comments">freezer</textarea>
-<fieldset>
- <legend> Pizza Size </legend>
- <p><input type=radio name=size value="small">Small</p>
- <p><input type=radio name=size value="medium" checked>Medium</p>
- <p><input type=radio name=size value="large">Large</p>
-</fieldset>
-<fieldset>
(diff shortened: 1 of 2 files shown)


tests/test_browser.py
```diff
@@ -74,14 +73,19 @@ def test__request():
     assert data["telephone"] == "555"
     assert data["comments"] == "freezer"
     assert data["size"] == "medium"
-    assert data["topping"] == ["cheese", "onion"]
+    assert data["topping"] == ["bacon", "onion"]
```
moy: This looks suspicious...

hemberger: This was simply to fix a flake8 error because the "Extra Cheese" line was too long, and removing "checked" made it shorter. My apologies for not documenting this change.

hemberger:
I see the cause of the coverage loss. I'll fix this and resubmit shortly.

tests/test_browser.py
hemberger: It is now documented in the revised commit message. Thanks for the reminder!

PR #219, opened 2018-07-12
Rewrite `Browser._request` to handle repeated input elements

hemberger:
Improve consistency of query string construction between
MechanicalSoup and web browsers in edge cases where form elements
have duplicate *name*-attributes.

For example, if a form with the following is submitted:

```
<input name="key" value="val1" />
<input name="key" value="val2" />
```

Then the resulting query string will now be

```
key=val1&key=val2
```

instead of

```
key=val2
```

More detail
===========

Store data to be encoded in the query string in a list of 2-tuples
instead of a dictionary. In doing so, we allow for the possibility
of multiple form elements being submitted with the same name.

Surprisingly, this is legal HTML (though sometimes unintentional),
so we want to be as faithful as possible to how the query string for
such a form would be created in a real browser.

To do this, we need to process the form elements in the order that
they appear on the page, rather than grouping them by type (which
is what we were doing when we stored the data in a dict).

This change allows us to accept any (name,value) combinations at
any point in the page without worrying about handling name conflicts
or whether to store a single value or a list of values -- because
now every bit of data is simply appended to the list of 2-tuples.

Ultimately, this means that the method is now agnostic to the
validity of the form data. While the `Form` class will still prevent
most illegal form manipulations, it has no control over HTML that
is malformed when it is served. There are some fringe cases where
malformed form HTML would cause MechanicalSoup to throw, but will
now submit the data as-is just like a browser would.

Note: We have simplified some of the logic by changing the bs4
selector to restrict to tags with name-attributes for all tag-types
(previously it only made this restriction for `input` and `button`,
even though it still ignored name-less `textarea` and `select`).

---------------------
This replaces PR's #159 and #160 with a, hopefully, more generic
solution to the issue. The `Browser._request` internals are still more
complicated than I'd like them to be, but I think this change at least
handles everything more consistently.

### DIFF
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -187,17 +187,19 @@ class Browser(object):
                 data.append((name, tag.text))
 
             elif tag.name == "select":
-                multiple = "multiple" in tag.attrs
-                values = []
-                for i, option in enumerate(tag.select("option")):
-                    if (i == 0 and not multiple) or "selected" in option.attrs:
-                        values.append(option.get("value", ""))
-                if multiple:
-                    for value in values:
+                options = tag.select("option")
+                selected_values = [i.get("value", "") for i in options
+                                   if "selected" in i.attrs]
+                if "multiple" in tag.attrs:
+                    for value in selected_values:
                         data.append((name, value))
-                elif values:
+                elif selected_values:
+                    # A standard select element only allows one option to be
+                    # selected, but browsers pick last if somehow multiple.
+                    data.append((name, selected_values[-1]))
+                elif options:
                     # Selects the first option if none are selected
-                    data.append((name, values[-1]))
+                    data.append((name, options[0].get("value", "")))
 
         if method.lower() == "get":
             kwargs["params"] = data
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -139,6 +139,7 @@ class Browser(object):
         return response
 
     def _request(self, form, url=None, **kwargs):
+        """Extract input data from the form to pass to a Requests session."""
         method = str(form.get("method", "get"))
         action = form.get("action")
         url = urllib.parse.urljoin(url, action)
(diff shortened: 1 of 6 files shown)


tests/test_form.py
```diff
@@ -235,9 +235,9 @@ def test_set_select_multiple(options):
     # When a browser submits multiple selections, the qsl looks like:
     #  name=option1&name=option2
     if not isinstance(options, list) and not isinstance(options, tuple):
-        expected = (('instrument', options),)
+        expected = [('instrument', options)]
```
moy: I agree that we need to test the sequence, not the set, but changing from tuple (`(...)`) to list (`[...]`) seems unnecessary (but not really harmful either, just that I like immutable things when I don't need to mutate them).

mechanicalsoup/browser.py
```diff
@@ -149,53 +150,54 @@ def _request(self, form, url=None, **kwargs):
...
+                    if not value:
+                        continue
+                    if isinstance(value, string_types):
+                        value = open(value, "rb")
+                    files[name] = value
+                else:
+                    data.append((name, value))
+
+            elif tag.name == "textarea":
+                data.append((name, tag.text))
```
moy: Ah, and this `if not name` is useless since you ignore nameless attributes in the CSS selector. Nice.

mechanicalsoup/browser.py
```diff
@@ -149,53 +150,54 @@ def _request(self, form, url=None, **kwargs):
...
+                multiple = "multiple" in tag.attrs
+                values = []
+                for i, option in enumerate(tag.select("option")):
+                    if (i == 0 and not multiple) or "selected" in option.attrs:
+                        values.append(option.get("value", ""))
+                if multiple:
+                    for value in values:
+                        data.append((name, value))
+                elif values:
+                    # Selects the first option if none are selected
```
moy: I think the comment is misplaced. It should be near the `if (i==0 and not multiple)` above. And actually, I think the code is a bit trickier than it needs to be. I'd just do something like

```
                options = tag.select("option")
                for i, option in enumerate(options):
                    if "selected" in option.attrs:
                        values.append(option.get("value", ""))
```

and then

```
                elif values:
                    data.append((name, values[-1]))
                elif options:
                    # Select the first option if none are selected
                    data.append((name, options[0].get("value", "")))
```

(Untested, but should be more self-explanatory and not much longer)

tests/test_form.py
hemberger: I agree that would probably be preferable, but since the output of `parse_qsl` is explicitly a list object, I saw only two sensible options:
1. Define the expected output as a list.
2. Convert the expected output to a list in `mock_post`.

I chose the first option since it seemed like the most direct comparison and it could not hide any errors behind a cast to another type. However, I'm open to alternatives if you see something better.

moy: It makes sense, indeed.

mechanicalsoup/browser.py
hemberger: I was quite confused by this code originally (not least because of the complete lack of comments), so I am very happy to rework it. I took your idea here and made a few modifications. How does this look to you?
```python
            elif tag.name == "select":
                options = tag.select("option")
                selected_values = [i.get("value", "") for i in options
                                   if "selected" in i.attrs]
                if "multiple" in tag.attrs:
                    for value in selected_values:
                        data.append((name, value))
                elif selected_values:
                    # A standard select element only allows one option to be
                    # selected, but browsers pick last if somehow multiple.
                    data.append((name, selected_values[-1]))
                elif options:
                    # Pick the first option if none are selected
                    data.append((name, options[0].get("value", "")))
```

mechanicalsoup/browser.py
```diff
@@ -149,53 +150,56 @@ def _request(self, form, url=None, **kwargs):
...
+                else:
+                    data.append((name, value))
+
+            elif tag.name == "textarea":
+                data.append((name, tag.text))
+
+            elif tag.name == "select":
+                options = tag.select("option")
+                selected_values = [i.get("value", "") for i in options
+                                   if "selected" in i.attrs]
```
moy: Indeed, nice use-case for a comprehension. Better than my naive `for` loop.

moy:
Hmm, I forgot another minor thing: there's no ChangeLog entry in this PR. You may want to re-use the one you wrote here:

https://github.com/MechanicalSoup/MechanicalSoup/pull/160/files#diff-4dc38e402d240900f2675401a08d2983

PR #224, opened 2018-08-03
conditionally require pytest-runner in setup_requires

graingert:


graingert:
this causes pytest-runner to be installed as part of the `egg` info and it bypasses pinned requirements from parent packages.

moy:
Can you include your comment in the commit message?

Thanks.

hemberger:
Does removing this prevent us from running tests with setup.py (e.g. `python setup.py test`)? Would it work to move `pytest-runner` into the `tests/requirements.txt` file? What we currently do is [explicitly recommended by pytest](https://docs.pytest.org/en/latest/goodpractices.html#integrating-with-setuptools-python-setup-py-test-pytest-runner), so I would like to understand this change a bit more before merging.

graingert:
Hmm. Maybe that should be fixed on the pytest docs.

If you released universal wheels it would also fix the problem

graingert:
https://github.com/MechanicalSoup/MechanicalSoup/issues/225

graingert:
@hemberger @moy I've updated my PR to match https://github.com/pytest-dev/pytest-runner#conditional-requirement

setup.py
```diff
@@ -1,8 +1,10 @@
-from setuptools import setup  # Always prefer setuptools over distutils
+import sys
 from codecs import open  # To use a consistent encoding
 from os import path
 import re
 
+from setuptools import setup  # Always prefer setuptools over distutils
+
```
moy: A detail: avoid committing unrelated changes (moving this `import` line has nothing to do with the conditional inclusion of pytest-runner).

setup.py
```diff
@@ -30,6 +32,9 @@ def read(fname, URL):
           'r', 'utf-8') as f:
     exec(f.read(), about)
 
+needs_pytest = {'pytest', 'test', 'ptr'}.intersection(sys.argv)
```
moy: IMHO this deserves a comment like

```
# Don't install pytest-runner on every setup.py run, just for tests.
# See https://pypi.org/project/pytest-runner/#conditional-requirement
```

setup.py
graingert: I moved it to import `sys`

graingert:
fyi, you don't run `python setup.py test` in CI, so it might be better to just remove `pytest-runner` completely

moy:
> fyi, you don't run python setup.py test in CI

No (I don't remember whether there's a good reason to run `pytest` over `python setup.py test` actually), but users may want to. One nice thing with setuptools is that the same commands are supposed to work in almost any Python projects, so a new contributor may want to run `python setup.py test` before even reading the doc.

setup.py
graingert: I've added this comment now

Issue #230, opened 2018-08-29
Action attribute not properly handled when selecting form


dhananjay-bajaj:
Hello sir,
When I select a form from browser object the target URL doesn't handle relative paths and raises an error as it can be seen in the screenshot provided below.
Error was produced when the following code was run:

<p><code>import mechanicalsoup

b=mechanicalsoup.StatefulBrowser()

r=b.open('https://google.co.in')

form=b.select_form(nr=0)

form.input({'q':'parrot'})

b.submit(form)
</code><p>

![mechanicalsoup_issue](https://user-images.githubusercontent.com/33071426/44778870-a6b59a00-ab9b-11e8-97b1-2e864f15991e.png)

PR #233, opened 2018-08-29
Override submit method in StatefulBrowser

hemberger:
stateful_browser.py: rename submit_selected to submit  …

The obvious method to use when trying to submit a form is the `submit`
method. Unless you are following the documentation explicitly, you
may overlook the `submit_selected` method (which is really the only
correct method to use when submitting a form within a StatefulBrowser
instance).

This is demonstrated by #230, where the user got an unexpected error
because they used `submit` instead of `submit_selected`.

To adhere to the principle of least astonishment, make `submit` a
synonym for `submit_selected` (with `submit` being the primary name
and `submit_selected` being an alias for backwards compatibility).

This is a breaking change for anyone calling the base class `submit`
from a StatefulBrowser instance. It is arguable that even if the call
is made correctly, it is the "wrong" way to submit, so we should make
it more difficult to "do the wrong thing".

You can still access the base class `Browser.submit` from an instance
of `StatefulBrowser` if you need to by using the `super` builtin, e.g.

```python
br = mechanicalsoup.StatefulBrowser()
super(StatefulBrowser, br).submit(...)
```

NOTE: I've also added a commit to improve backwards compatibility and
issue a deprecation warning if a user tries to call `Browser.submit`
from a StatefulBrowser instance _if_ they are passing a Form instance
(we can't catch cases where they are passing a bs4.element.Tag -- see
the commit message for details).

### DIFF
--- a/mechanicalsoup/stateful_browser.py
+++ b/mechanicalsoup/stateful_browser.py
@@ -7,6 +7,7 @@ from .form import Form
 import sys
 import re
 import bs4
+import warnings
 
 
 class _BrowserState:
@@ -220,6 +221,19 @@ class StatefulBrowser(Browser):
         to :func:`Form.choose_submit` on the current form to choose between
         them. All other arguments are forwarded to :func:`Browser.submit`.
         """
+        # Temporarily allow calling the old inherited Browser.submit directly
+        # in case we can detect with certainty that the call is an old style.
+        # Note: Browser.submit also accepts a bs4.element.Tag with name="form",
+        # but we cannot assume this is an old-style call since there could be
+        # a submit button with name="form".
+        if isinstance(btnName, Form):
+            warnings.warn("This usage of StatefulBrowser.submit is deprecated."
+                          " Please see the documentation for this function to "
+                          "upgrade to the new interface.",
+                          DeprecationWarning)
+            return super(StatefulBrowser, self).submit(btnName, *args,
+                                                       **kwargs)
+
         self.get_current_form().choose_submit(btnName)
 
         referer = self.get_url()
--- a/mechanicalsoup/stateful_browser.py
+++ b/mechanicalsoup/stateful_browser.py
@@ -60,6 +60,10 @@ class StatefulBrowser(Browser):
         self.__verbose = 0
         self.__state = _BrowserState()
 
+        # Aliases for backwards compatibility
+        # (Included specifically in __init__ to suppress them in Sphinx docs)
+        self.submit_selected = self.submit
+
     def set_debug(self, debug):
         """Set the debug mode (off by default).
 
@@ -207,7 +211,7 @@ class StatefulBrowser(Browser):
 
         return self.get_current_form()
 
-    def submit_selected(self, btnName=None, *args, **kwargs):
+    def submit(self, btnName=None, *args, **kwargs):
         """Submit the form that was selected with :func:`select_form`.
 
         :return: Forwarded from :func:`Browser.submit`.
@@ -225,8 +229,9 @@ class StatefulBrowser(Browser):
             else:
                 kwargs['headers'] = {'Referer': referer}
 
-        resp = self.submit(self.__state.form, url=self.__state.url,
-                           *args, **kwargs)
+        resp = super(StatefulBrowser, self).submit(self.__state.form,
+                                                   url=self.__state.url,
+                                                   *args, **kwargs)
         self.__state = _BrowserState(page=resp.soup, url=resp.url,
                                      request=resp.request)
         return resp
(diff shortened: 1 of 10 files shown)


mechanicalsoup/stateful_browser.py
```diff
@@ -216,6 +221,19 @@ def submit_selected(self, btnName=None, *args, **kwargs):
...
+        # Temporarily allow calling the old inherited Browser.submit directly
+        # in case we can detect with certainty that the call is an old style.
+        # Note: Browser.submit also accepts a bs4.element.Tag with name="form",
+        # but we cannot assume this is an old-style call since there could be
+        # a submit button with name="form".
+        if isinstance(btnName, Form):
+            warnings.warn("This usage of StatefulBrowser.submit is deprecated."
+                          " Please see the documentation for this function to "
+                          "upgrade to the new interface.",
+                          DeprecationWarning)
```
moy: Unfortunately, forbidding this is not only a backward-compatibility problem, but also a violation of Liskov's Substitution Principle. The base `.submit()` accepts a `btnName` instance of `Form`, so all derived methods should also accept at least that (and possibly more), so that anything doable with a `Browser` is also doable with a `StatefulBrowser`.

Also, having the same positional argument be `btnName` in one class and `form` in another is very misleading. LGTM is right to complain here IMHO.

moy (changes requested):

mechanicalsoup/stateful_browser.py
hemberger: Before I invest too much more time into this, I just wanted to check if you are generally opposed to this PR, or if you can imagine a modified PR that you would consider accepting. Even better if you already have an idea in mind! ;)

My preference is always to make `StatefulBrowser` the best class it can be. The `Browser` class sometimes gets in the way of that, such as in this case. I believe `StatefulBrowser` would benefit from a `submit` method, so if we can't override `Browser.submit` in the proposed way, the first two alternatives that come to mind (setting aside the issue of backwards compatibility for a moment) are renaming `Browser.submit` to either a) `Browser._submit` or b) `Browser.submit_form`.

moy: If I were to redo this from scratch, I'd consider using composition instead of inheritance, i.e. consider that a `StatefulBrowser` *contains* a `Browser`, not that it *is* a `Browser`. It'd be hard to change that without breaking backward compatibility, though.

Renaming `Browser.submit` is also problematic with respect to backward compatibility. Perhaps not many people call `submit` on a `StatefulBrowser`, but everyone who wrote code before `StatefulBrowser` was added does call `submit` on a `Browser`.

There's another option: add warnings for suspicious cases, and allow the user to disable the warnings as needed. We could warn on calls to `StatefulBrowser.submit`, or if we want to go further, on creation of a `Browser` that isn't a `StatefulBrowser`. Probably not as a deprecation warning like "what you're doing won't work anymore soon", but just like "uh, are you sure you want that?".

hemberger: A warning for suspicious use is a great idea, but I'm not sure that it addresses the main issues here (and might be just as much of an annoyance as a modified interface to some users). Out of curiosity, though, did you have a design in mind for warning toggling?

Okay, two more possibilities:
1. Simply add a comment to the `Browser.submit` docstring saying something like:
> If you are calling this method from a `StatefulBrowser` instance, consider using `StatefulBrowser.submit_selected` instead.

2. Override `Browser.submit` in a Liskov-compliant way, e.g.
```python
def submit(self, form=None, url=None, **kwargs, btnName=None):
```
where `form` and `url` default to the current form/url if `None`. It's a bit of a clunky interface, but might be the best we can do if we want to make overriding work within our constraints. Or even just
```python
def submit(self, form=None, url=None, **kwargs):
```
to avoid complications with `btnName`.

hemberger: @moy I'd like to get version 0.11 released as soon as possible, but I want to address this issue in some way before I do, even if it's just a minor addition to the documentation.

How about I make the documentation change proposed above for 0.11, and then we can reassess if we want to make any code changes for version 1.0 later?

moy: The doc change is obviously a good step forward.

Your option 2. is probably doable too. I won't have time to look at this in details soon, but if you're confident enough I trust you to do the right thing.

hemberger: Thanks for the feedback! For the 0.11 release I'll just change the docstring, unless you recommend otherwise.

As I think more about it, overriding `submit` seems increasingly important. One of the biggest issues (that I didn't appreciate until just now) is that if you call `submit` on a StatefulBrowser, the browser state becomes stale. I would argue that this is a pretty dangerous inconsistency for a method that looks like the most obvious way to submit a form.

dillonko:
@moy With regards to how old this PR is and whether it's a valid issue any more should this be closed?

PR #246, opened 2018-10-13
Case insensitive type="radio" and type="checkbox"

### LINKED ISSUE #245: Tag type maching case sensitive
When the form uses upper casing the matching would fail.

https://github.com/MechanicalSoup/MechanicalSoup/blob/5926d0efd887097f9da56241ae945c4d5a81d8ba/mechanicalsoup/browser.py#L166

Suggest adding .lower()
`if tag.get("type").lower() in ("radio", "checkbox"):`

moy:
Fixes #245. It was actually a bit more involved than just adding a `.lower()` in the code, since there are several places where we look for radio and checkboxes.

### DIFF
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -163,7 +163,10 @@ class Browser(object):
             name = tag.get("name")  # name-attribute of tag
 
             if tag.name == "input":
-                if tag.get("type") in ("radio", "checkbox"):
+                if (
+                        tag.get("type") is not None and
+                        tag.get("type").lower() in ("radio", "checkbox")
+                ):
                     if "checked" not in tag.attrs:
                         continue
                     value = tag.get("value", "on")
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -174,7 +174,10 @@ class Browser(object):
                     # browsers use empty string for inputs with missing values
                     value = tag.get("value", "")
 
-                if tag.get("type") == "file":
+                if (
+                        tag.get("type") is not None and
+                        tag.get("type").lower() == "file"
+                ):
                     # read http://www.cs.tut.fi/~jkorpela/forms/file.html
                     # in browsers, file upload only happens if the form
                     # (or submit button) enctype attribute is set to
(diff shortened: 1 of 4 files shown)


Pimmetje:
I could agree that this isn't be the best place to fix it. Now i know the issue i can work around it in my code. 
The question is what is a good case to report on bs4.element.Tag that would fix the issue?

moy:
> There are a lot of other places than just "radio" and "checkbox" where we perform case-sensitive string comparison of the type value. To do this for only a subset of the types would be inconsistent.

I didn't do a very thorough inspection, but a quick grep for `type` shows that there are not _that_ many instances. I've added one commit dealing with a few other cases. I may have missed more cases, but it seems to be complete. In its current shape the patch does serve a purpose, I think it's sufficient as-is and tolerance to non-lowercase type attributes can be improved later if we find more corner-cases where it doesn't work.

> I'm surprised BeautifulSoup doesn't handle this issue somehow. It explicitly states that it converts the tags and attribute keys to lowercase (https://www.crummy.com/software/BeautifulSoup/bs4/doc/index.html?highlight=tag#other-parser-problems), so it is at least tangentially aware of this problem.

For tags, it's very clear that they are case-insensitive. For attribute names, they can contain more or less arbitrary data and it would be wrong from BeautifulSoup to modify this data (think of what would happen if you lower-cased all `value=` attributes ...). I was tempted to lower-case all attributes when creating the `Form` object, but I believe this would be wrong too for the same reasons. Lower-casing only `type=` attributes could work, but feels like a hack too. I didn't find a definite reference for that but I believe this `type="RaDiO"` is not proper HTML (and if so, this gives one more argument for point 1.: if it's incorrect HTML I'm much less hesitant to apply a best effort approach).

> This implementation might be difficult to maintain. If you add new branch, you have to remember that you're supposed to make case-insensitive type comparisons. Maybe there's a way to do this at a lower level?

Well, I'd love a `case_insensitive=True` argument to BeautifulSoup's search methods, but apparently they are not there. I've added a small helper function to do this in `Form`, which makes the code a bit more compact. Obviously, one still has to remember to use it.

hemberger:
I'm sorry I haven't had the chance to review this again, but I hope to take a look at it again soon. If I'm holding things up, please go ahead and merge. Otherwise I'd like to think about this some more. Thanks!

moy:
I had a quick look again, and it looks good to me. I had forgotten about all this and was convinced by my own arguments while reading the thread ;-).

hemberger:
Ah, I see I missed the boat here. Sorry for being so slow about this. :(

PR #247, opened 2018-10-18
Update expl_duck_duck_go.py

pbk000:


hemberger:
That's cute :)

I noticed you made a bunch of little easter egg PR's. What inspired them?

Issue #248, opened 2018-11-11
Do not submit disabled <input> elements


5j9:
https://www.w3.org/TR/html52/sec-forms.html#element-attrdef-disabledformelements-disabled
> The disabled attribute is used to make the control non-interactive and to prevent its value from being submitted.

MechanicalSoup ignores ``disabled`` attributes which should be fixed.

Some additional notes: (from https://www.wufoo.com/html5/disabled-attribute/)

> * If the disabled attribute is set on a ``<fieldset>``, the descendent form controls are disabled.
> * A disabled field can’t be modified, tabbed to, highlighted, or have its contents copied. Its value is also ignored when the form goes thru constraint validation.
> * The disabled value is Boolean, and therefore doesn’t need a value. But, if you must, you can include disabled="disabled".
> * Setting the value of the disabled attribute to null does not remove the effects of the attribute. Instead use removeAttribute('disabled').
> * You can target elements that are disabled with the :disabled pseudo-class. Or, if you want to specifically target the presence of the attribute, you can use input[disabled]. Similarly, you can use :enabled and input:not([disabled]) to target elements that are not disabled.
> * You do not need to include aria-disabled="true" when including the disabled attribute because disabled is already well supported. However, if you are programmatically disabling an element that is not a form control and therefore the disabled attribute does not apply, include aria-disabled="true".
> * The disabled attribute is valid for all form controls including all ``<input>`` types, ``<textarea>``, ``<button>``, ``<select>``, ``<fieldset>``, and ``<keygen>``.
>

hemberger:
Indeed, it looks like we're not dealing with this appropriately in MechanicalSoup. It will take some work to make us compliant with all the details of `disabled` attributes, but I think as a first pass we can easily add it into the `Browser._request` method when we process the form data.

If you'd like to submit a PR, we would happily welcome it and can provide any guidance that you need. Otherwise, we'll get to this as soon as we can.

Thanks for submitting this issue!

rationa-kunal:
I would like to work on this issue !

moy:
@rationa-kunal: thanks! Let us know if you need help, and don't hesitate to post drafts if you want feedback.

rationa-kunal:
I was thinking of adding an extra boolean variable in Form class as dev_option. If that is set to True then we can edit disabled values and if it is False we cant edit it.
for example .
```python
if i.has_attr("disabled") and not self.dev_op:
[...]

PR #254, opened 2018-12-03
added ability to not edit disabled

rationa-kunal:
Added dev_op boolean variable to Form
if dev_op is True then you can edit disabled inputs
else dev_op is False then you cant edit disabled inputs

### DIFF
--- a/mechanicalsoup/form.py
+++ b/mechanicalsoup/form.py
@@ -33,7 +33,7 @@ class Form(object):
     It also handles submit-type elements using :func:`~Form.choose_submit`.
     """
 
-    def __init__(self, form):
+    def __init__(self, form, dev_op=False):
         if form.name != 'form':
             warnings.warn(
                 "Constructed a Form from a '{}' instead of a 'form' element. "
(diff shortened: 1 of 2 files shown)


moy:
We have automatic testing, and as you see there are issues to fix in your code (at least flake8 formatting issues). This would also require documentation and more tests (we have a strict 100% coverage policy).

But as I said in #248, I don't think this is the right approach. If you think it is, then please rewrite the commit message to explain 1) the issue you are trying to solve and 2) why you think your modifications are the best way to fix this issue.

moy:
As I wrote in https://github.com/MechanicalSoup/MechanicalSoup/issues/248#issuecomment-443687132, I don't think edition of disabled fields is the problem. The first thing to fix is the form _submission_.

Also, the Travis build fails, we can't merge a PR with test failures, sorry.

rationa-kunal:
How can I add test case for this issue ?
There is no disabled form field in httpbin !

moy:
I believe 685eb6e1396008a0b3881bea5bcc6b7119f61642 fixed the main problem with disabled fields, so let's close this.

PR #261, opened 2019-01-17
Improve `test_link_arg_*` tests

hemberger:
The `test_link_arg_regex` test was accidentally passing. By adding
another link to the test html, we can see that it was selecting the
first link instead of the link that matched the supplied regex.
This new version of the test fails, but it will be fixed by #256.

Also added a case for no arguments and reduced code duplication
by changing it to a parameterized test.

I'm not sure what way would be best to merge this in, since it should
come before #256 (to demonstrate the existing failure), but we probably
don't want the master branch to be failing for an extended period of time.
Any thoughts?

### DIFF
--- a/tests/test_stateful_browser.py
+++ b/tests/test_stateful_browser.py
@@ -421,18 +421,18 @@ def test_referer_submit_headers(httpbin):
     assert headers['X-Test-Header'] == 'x-test-value'
 
 
-def test_link_arg_text(httpbin):
+@pytest.mark.parametrize('expected, kwargs', [
+    pytest.param('/foo', {}, id='none'),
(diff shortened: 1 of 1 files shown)


moy:
Even if it doesn't happen for a long period of time, I prefer to avoid temporary breakage. This makes history bisectable and IMHO more readable.

You can mark the test as known failure with `@pytest.mark.xfail` in this commit, and the removal of this mark in the next commit documents the fact that it fixes the bug.

hemberger:
Aha, I had forgotten about `pytest.mark.xfail`. Thanks for the suggestion! Should be fixed now.

PR #263, opened 2019-01-17
form.py: replace find_by_type with case-insensitive selectors

hemberger:
This change requires bs4 4.7.0+ (since older versions of bs4 do not
follow the CSS selector spec as closely as SoupSieve does).

I think we should delay applying this patch, since a hard dependency
on bs4 4.7.0 is too restrictive right now, but I do think that the
resulting code is more compact and easier to read.

### DIFF
--- a/mechanicalsoup/form.py
+++ b/mechanicalsoup/form.py
@@ -111,7 +111,8 @@ class Form:
         """
         for (name, value) in data.items():
             # Case-insensitive search for type=checkbox
-            checkboxes = self.find_by_type("input", "checkbox", {'name': name})
+            selector = 'input[type="checkbox" i][name="{}"]'.format(name)
+            checkboxes = self.form.select(selector)
             if not checkboxes:
                 raise InvalidFormMethod("No input checkbox named " + name)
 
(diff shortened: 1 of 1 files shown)


hemberger:
Welcome back @facelessuser !

In my testing, I don't see SoupSieve (or at least bs4) treating `type` attributes as case-insensitive by default. Am I doing something silly here?
```
In [10]: bs4.BeautifulSoup('<input type="CHECKBOX" />').select('input[type="checkbox"]')                                                      
Out[10]: []

In [11]: bs4.BeautifulSoup('<input type="CHECKBOX" />').select('input[type="checkbox" i]')                                                    
Out[11]: [<input type="CHECKBOX"/>]
```
I certainly wouldn't object to SoupSieve handling the case-sensitivity issue of `type` so that we don't have to ;)

facelessuser:
It was a bug. I was testing it against HTML5. We were checking if the document was under the html namespace, but that only applies to HTML5 and XHTML. When using the default parser or the `lxml` html parser, there is no namespace applied to the document. And this is why I'm poking around 🙂 .

facelessuser:
I have a fix for `type` selectors here: https://github.com/facelessuser/soupsieve/pull/88.

facelessuser:
Soup Sieve 1.7.2 has been released and fixes the insensitive issue. To avoid these issues in the future, we now iterate through all the parsers on each test (except where it doesn't make sense).  As this issue is deferred for you, you should have what you need when you get around to actually including the changes. But you are always free to use the `i` flag regardless.

Issue #264, opened 2019-01-24
Cannot submit form with multiple identical submits


ricecooker:
When I have a form like:

```
<form action="#" method="POST" name="blah">
  <input type="submit" name="search" value="Search"/>
  <input type="submit" name="search" value="Search"/>
</form>
```
There isn't a way to submit the form. It will *always* raise a `LinkNotFoundError`.

Around form.py:333 I'd do 
```
if inp == submit and found: 
  del inp['name']
```

amrit94:
Please add button name  in the HTML like:
```
<input type="submit" value="Submit" name ='button1'>
<input type="submit" value="Submit" name ='button2'>
```
And in mechanicalsoup submit using btnName = 'buttonName'

> response = browser.submit_selected(btnName='button1')

nk9:
Thanks for fixing this. The last release was from a year ago, though. If I need this now, what's the recommended branch? Just… master? ***gulp***

hemberger:
I'm sure we could come up with a new release on fairly short order. I'll talk to @moy and we'll keep you posted.

nk9:
That would be swell, thanks.

For others who may be running into the problem and are using v0.11.0 or earlier, I have found a simple solution:

```
submitEls = browser.get_current_page().find_all(type='submit')

# Remove excess submit elements
for index, submitEl in enumerate(submitEls):
	if index > 0:
		submitEl.extract()
```

PR #265, opened 2019-01-24
Fixing issue of duplicate submit buttons

ricecooker:
Suggested fix for #264

I broke up the `if` statement to separate the logic of matching by name attribute string versus matching by `Tag`.

### DIFF
--- a/mechanicalsoup/form.py
+++ b/mechanicalsoup/form.py
@@ -344,12 +344,17 @@ class Form(object):
 
         found = False
         for inp in inps:
-            if inp == submit or (inp.has_attr('name') and
-                                 inp['name'] == submit):
+            if (inp.has_attr('name') and inp['name'] == submit):
                 if found:
                     raise LinkNotFoundError(
                         "Multiple submit elements match: {0}".format(submit)
                     )
+                found = True
+            # Ignore submit element since it is an exact duplicate of
+            # the one we're looking at.
+            elif inp == submit:
+                if found:
+                    del inp['name']
                 found = True
             else:
                 # Delete any non-matching element's name so that it will be
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -467,6 +467,26 @@ def test_issue158():
     browser.close()
 
 
+def test_duplicate_submit_buttons():
+    """Tests that duplicate submits doesn't break form submissions
+    See issue https://github.com/MechanicalSoup/MechanicalSoup/issues/264"""
+    issue264_form = '''
+<form method="post" action="mock://form.com/post">
+  <input name="box" type="hidden" value="1"/>
+  <input name="search" type="submit" value="Search"/>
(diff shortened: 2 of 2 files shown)


moy:
I didn't look closely at the code but the patch looks reasonable. This lacks tests however: first, you're fixing a bug so you should add a test demonstrating the failure, and we have a strict 100% coverage policy.

Thanks

ricecooker:
I added a test for the bug so coverage should be 100%. I ran the test against `master` and the test failed because a LinkNotFoundError was raised because of duplicates.

tests/test_form.py
```diff
@@ -467,6 +467,25 @@ def test_issue158():
     browser.close()
 
 
+def test_issue264():
```
moy: I would prefer more meaningful test name (so that one can immediately know which test is about what when skimming through a list of tests, without having to open the corresponding issue each time) and a link to the issue in the docstring, but i'm OK with this if you prefer.

mechanicalsoup/form.py
```diff
@@ -344,13 +344,16 @@ def choose_submit(self, submit):
...
-            if inp == submit or (inp.has_attr('name') and
-                                 inp['name'] == submit):
+            if (inp.has_attr('name') and inp['name'] == submit):
                 if found:
                     raise LinkNotFoundError(
                         "Multiple submit elements match: {0}".format(submit)
                     )
                 found = True
+            elif inp == submit:
+                if found:
```
moy: This would deserve a comment, like

```
# The submit element is an exact duplicate of the one we're looking for, just ignore it
```

tests/test_form.py
ricecooker: Renamed to `test_duplicate_submit_buttons`

mechanicalsoup/form.py
ricecooker: Added your suggested comment

ricecooker:
> On overall, this looks good to me (see one minor point about test naming and another about a comment).
> 
> Two minor issues before we can merge:
> 
> * This is a 2-commits PR, with one commit failing continuous integration. Each commit should pass. In this case, I see no reason to keep commits separate, so squashing them together is probably the simplest.
> * The commit messages do not explain the problem, nor justify why the patch is a good solution to the problem.
> 
> You can address both points by using `git rebase -i`, squashing the commits and use `r` to reword the commit message (or `git commit --amend`).

Squashed into one commit.

moy:
Thanks. This still lacks a proper color message though.

ricecooker:
bump. i added more color in the commit message. please let me know if there's anything more to be done.

moy:
Perfect thanks (and LOL about the "color", I probably typed this message from my phone with a "smart" keyboard ...). I moved the comment explaining why the item was ignored to within the corresponding `if` statement while applying, and merged as 10baed43148868dd6794e7b2e65db994031550cc.

Issue #201, opened 2018-03-11
Handle image submits


mjpieters:
An `input` element of [type `image`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/image) is also a valid submit button:

```
<form><input type="text" name="search" /><input type="image" name="search" /></form>
```

Clicking on the image button will submit the form, as does hitting enter. The form then has two fields, `image.x` and `image.y`, set to the coordinates where the image was clicked, or `0` when hitting enter.

MechanicalSoup has no facilities for handling this type of input; it is not recognised as a valid submit and on submitting, no `.x` and `.y` entries are generated.

moy:
Indeed. Supporting `image` as a synonym for `submit` is straightforward. Support for `.x` and `.y` fields needs a bit of thinking: what should the MechanicalSoup API look like?

Perhaps

    br["image.x"] = x
    br["image.y"] = y
    br.submit_selected()

?

Then `br["image.x"]` would just check that there's an image input with name `image`, and store somewhere the value `x`. The easy implementation would be to add a hidden field to store the value. Perhaps it'd be nicer to store the value as part of the `image` tag, adding fields like `data-x` and `data-y`?

PR #268, opened 2019-02-07
Add support to input type "image"

matheuscas:
This PR intents to add support to inputs with type "image". Currently, Form only seeks "submit inputs" and buttons. I've added the search of "image submits". 

This PR is also based on this [issue](https://github.com/MechanicalSoup/MechanicalSoup/issues/201).

### DIFF
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -128,9 +128,10 @@ def test_choose_submit_from_selector_submit(value):
     pytest.param('cancel', id='second'),
 ])
 def test_choose_submit_from_selector_image(value):
-    """Test choose_submit by passing a CSS selector argument for a input type image.
-    This test aims to ensure that inputs with type image are found, regardless its repetition.
-    Based in part of this issue: https://github.com/MechanicalSoup/MechanicalSoup/issues/201
+    """Test choose_submit by passing a CSS selector arg for a input type image.
+    This test aims to ensure that inputs with type image are found,
+    regardless its repetition.
+    Issue: https://github.com/MechanicalSoup/MechanicalSoup/issues/201
     """
     text = """
     <form method="post" action="mock://form.com/post">
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -128,7 +128,10 @@ def test_choose_submit_from_selector_submit(value):
     pytest.param('cancel', id='second'),
 ])
 def test_choose_submit_from_selector_image(value):
-    """Test choose_submit by passing a CSS selector argument."""
+    """Test choose_submit by passing a CSS selector argument for a input type image.
+    This test aims to ensure that inputs with type image are found, regardless its repetition.
+    Based in part of this issue: https://github.com/MechanicalSoup/MechanicalSoup/issues/201
+    """
     text = """
     <form method="post" action="mock://form.com/post">
       <input type="image" name="do" value="continue" />
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -106,12 +106,33 @@ def test_choose_submit(expected_post):
     pytest.param('continue', id='first'),
     pytest.param('cancel', id='second'),
 ])
-def test_choose_submit_from_selector(value):
+def test_choose_submit_from_selector_submit(value):
     """Test choose_submit by passing a CSS selector argument."""
     text = """
     <form method="post" action="mock://form.com/post">
       <input type="submit" name="do" value="continue" />
       <input type="submit" name="do" value="cancel" />
+    </form>"""
+    browser, url = setup_mock_browser(expected_post=[('do', value)], text=text)
+    browser.open(url)
+    form = browser.select_form()
+    submits = form.form.select('input[value="{}"]'.format(value))
+    assert len(submits) == 1
+    form.choose_submit(submits[0])
+    res = browser.submit_selected()
+    assert res.status_code == 200 and res.text == 'Success!'
+
+
+@pytest.mark.parametrize("value", [
+    pytest.param('continue', id='first'),
+    pytest.param('cancel', id='second'),
+])
+def test_choose_submit_from_selector_image(value):
+    """Test choose_submit by passing a CSS selector argument."""
+    text = """
+    <form method="post" action="mock://form.com/post">
+      <input type="image" name="do" value="continue" />
+      <input type="image" name="do" value="cancel" />
     </form>"""
     browser, url = setup_mock_browser(expected_post=[('do', value)], text=text)
     browser.open(url)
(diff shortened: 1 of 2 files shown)


moy:
Thanks for your PR. Minor comments about the commit message:

* Avoid past tens ("Added ..."). We usually use imperative tone ("Add ..." as in "Please, dear computer, add feature ...")

* Include the string "Fixes #201" in the commit message. This way, the issue will be closed when the PR is merged, and the issues will be cross-linked.

tests/test_form.py
```diff
@@ -106,7 +106,7 @@ def test_choose_submit(expected_post):
     pytest.param('continue', id='first'),
     pytest.param('cancel', id='second'),
 ])
-def test_choose_submit_from_selector(value):
+def test_choose_submit_from_selector_submit(value):
```
moy: Why this renaming?

matheuscas: Because I basically replicated this same test, but for "image" type.

moy: Ah, OK.

matheuscas:
> The `.x` and `.y` attributes should ideally be set on submission, and they currently are not.
> 
> I'm OK with merging a version that doesn't support `.x` and `.y`, but it would be nice to have tests showing the current behavior (with proper comments explaining why it is like this). At least, the commit message should explain it.

I didn't quite get the `.x` and `.y` thing. That's why I focused only on the image attribute type.

matheuscas:
> Thanks for your PR. Minor comments about the commit message:
> 
> * Avoid past tens ("Added ..."). We usually use imperative tone ("Add ..." as in "Please, dear computer, add feature ...")
> * Include the string "Fixes #201" in the commit message. This way, the issue will be closed when the PR is merged, and the issues will be cross-linked.

Should I change anything here?

moy:
> I didn't quite get the `.x` and `.y` thing. That's why I focused only on the image attribute type.

They tell you where the mouse clicked, which only half makes sense for us because we don't have a mouse, but it would be cool to have a way to specify it.

See e.g. https://mdn.github.io/learning-area/html/forms/image-type-example/xy-coordinates-example.html

moy:
> > Thanks for your PR. Minor comments about the commit message:
> > 
> > * Avoid past tens ("Added ..."). We usually use imperative tone ("Add ..." as in "Please, dear computer, add feature ...")
> > * Include the string "Fixes #201" in the commit message. This way, the issue will be closed when the PR is merged, and the issues will be cross-linked.
> 
> Should I change anything here?

If you can modify your commit message (`git commit --amend`) to address both points and then force-push to your branch, that would be perfect.

tests/test_form.py
```diff
@@ -123,6 +123,27 @@ def test_choose_submit_from_selector(value):
...
+@pytest.mark.parametrize("value", [
+    pytest.param('continue', id='first'),
+    pytest.param('cancel', id='second'),
+])
+def test_choose_submit_from_selector_image(value):
+    """Test choose_submit by passing a CSS selector argument."""
+    text = """
+    <form method="post" action="mock://form.com/post">
+      <input type="image" name="do" value="continue" />
+      <input type="image" name="do" value="cancel" />
```
hemberger: > `<input type="image">` elements do not accept value attributes. The path to the image to be displayed is specified in the src attribute.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/image

While this obviously this impacts the test (which can be easily modified), it's not clear to me whether or not it impacts the implementation of the fix. I don't think it does, but it's something to think twice about.

tests/test_form.py
```diff
@@ -123,6 +123,27 @@ def test_choose_submit_from_selector(value):
     assert res.status_code == 200 and res.text == 'Success!'
 
 
+@pytest.mark.parametrize("value", [
+    pytest.param('continue', id='first'),
+    pytest.param('cancel', id='second'),
+])
+def test_choose_submit_from_selector_image(value):
+    """Test choose_submit by passing a CSS selector argument."""
```
hemberger: We're all guilty of writing docstrings that are not sufficiently detailed, but for the sake of this as a learning experience, I'll add that it would be nice if you could make the docstring say more specifically what the test is designed to do (you can even reference an Issue number here if you like).

matheuscas: Take a look.

hemberger: Looks great, thanks for the extra detail!

tests/test_form.py
matheuscas: I think I got what u mean. This test could be an input type submit and another input type image and making sure that the input image would be picked - with different names, of course. What do you think?

hemberger: Ultimately, we need to try to replicate what the browser does (especially in cases where this is the standard). When clicking on a `<input type="image" name="test" ...>`, the browser does not submit a `name=value` request, because `value` is an invalid attribute for `type="image"`. Instead, it will submit a request that looks like:
```
test.x=171&test.y=168 
```
where 171 and 168 are the mouse positions where the image was clicked. Obviously we don't have access to mouse positions with MechanicalSoup, so I'm not exactly sure how we _should_ handle this, but I know we don't want to try to use a `value` in this case (or at the very least, we should allow for the case where `value` doesn't exist, and just submit a `name=` request). Or maybe we should submit `name.x=0&name.y=0`?

@moy Do you have any thoughts about what the request should look like?

moy: I think `name.x=0&name.y=0` by default makes sense. Ideally, there would be a way to specify `x` and `y` from MechanicalSoup, but this can wait for another PR.

hartas17:
Hello is this viable for input like this?

```
<input border="0" class="pdf" id="ctl00_ctl02_DescargaPDF" name="ctl00$ctl02$DescargaPDF" src="imagenes/icono_pdf.png" title="Descarga Pdf" type="image"/>
<input border="0" class="pdf" id="ctl00_ctl03_DescargaPDF" name="ctl00$ctl03$DescargaPDF" src="imagenes/icono_pdf.png" title="Descarga Pdf" type="image"/>
<input border="0" class="pdf" id="ctl00_ctl04_DescargaPDF" name="ctl00$ctl04$DescargaPDF" src="imagenes/icono_pdf.png" title="Descarga Pdf" type="image"/>
<input border="0" class="pdf" id="ctl00_ctl05_DescargaPDF" name="ctl00$ctl05$DescargaPDF" src="imagenes/icono_pdf.png" title="Descarga Pdf" type="image"/>

```
my code is:

`browser.open(url)
`
`form = browser.select_form()
`
`inputs = browser.get_current_page().find_all('input',title="Descarga Xml")
`
`input_to_download = inputs[0]
`
`name_input = input_to_download.attrs.get('name',None)
`
`form.choose_submit(name_input)
`
`text = browser.submit_selected()
`

This input should return a file PDF.

 If i make a request with **postman interceptor** is like:
`[{"key":"ctl00$PHContenidoPag$ddl","value":"827731000222"},{"key":"ctl00$ctl02$DescargaPDF.x","value":"9"},{"key":"ctl00$ctl02$DescargaPDF.y","value":"6"}]`

`ctl00$PHContenidoPag$ddl` is value from a selected value in a <select form

PR #276, opened 2019-03-29
Respecting enctype specified

### LINKED ISSUE #242: Respect enctype specified in forms
MechanicalSoup currently ignores the enctype specified for a form when submitting it, instead using the default used by requests (which is application/x-www-form-urlencoded unless you specify "files" to upload).  I'd like it to behave more like a browser by respecting the given enctype (apparently you should specify "files" to requests to upload form fields that aren't files as multipart/form-data - which is the case I'm concerned with, making sure to submit as multipart/form-data when that's what the form specifies as its enctype).  This is for testing web applications which behave differently when you submit with the wrong enctype (concretely, the Python cgi module defaults to ignoring fields with empty values for application/x-www-form-urlencoded but not for multipart/form-data, so any application using that may behave differently for submissions with the wrong enctype).

MartinDlry:
Fixes #242

### DIFF
--- a/tests/test_browser.py
+++ b/tests/test_browser.py
@@ -81,55 +81,95 @@ def test__request(httpbin):
         "Content-Type"]
 
 
-@pytest.mark.parametrize('expected_content, set_value', [
-    (b":-)", True),
-    (b"", False)
+valid_enctypes_file_submit = {"multipart/form-data": True,
+                              "application/x-www-form-urlencoded": False
+                              }
+
+default_enctype = "application/x-www-form-urlencoded"
+
+# create a temporary file for testing file upload
+file_content = b":-)"
+pic_filedescriptor, pic_path = tempfile.mkstemp()
+os.write(pic_filedescriptor, file_content)
+os.close(pic_filedescriptor)
+
+# key is :
+# valid_enctype_file_submit[enctype], file_submit
+expected_content = {
+    (False, False): (b"", "form"),
+    (False, True): (pic_path.encode(), "form"),
+    (True, False): (b"", "files"),
+    (True, True): (file_content, "files")
+}
+
+
+@pytest.mark.parametrize("file_field", [
+  """<input name="pic" type="file" />""",
+  ""])
+@pytest.mark.parametrize("submit_file", [
+    True,
+    False
 ])
-def test__request_file(httpbin, expected_content, set_value):
+@pytest.mark.parametrize("enctype", [
+  pytest.param("multipart/form-data"),
+  pytest.param("application/x-www-form-urlencoded"),
+  pytest.param("Invalid encytpe")
+])
+def test_enctype_and_file_submit(httpbin, enctype, submit_file, file_field):
+    # test if enctype is respected when specified
+    # and if files are processed correctly
     form_html = """
-    <form method="post" action="{}/post">
-      <input name="pic" type="file" />
+    <form method="post" action="{}/post" enctype="{}">
+      <input name="in" value="test" />
+      {}
     </form>
-    """.format(httpbin.url)
+    """.format(httpbin.url, enctype, file_field)
     form = BeautifulSoup(form_html, "lxml").form
 
-    if set_value:
-        # create a temporary file for testing file upload
-        pic_filedescriptor, pic_path = tempfile.mkstemp()
-        os.write(pic_filedescriptor, expected_content)
-        os.close(pic_filedescriptor)
-
+    if submit_file and file_field:
         form.find("input", {"name": "pic"})["value"] = pic_path
 
     browser = mechanicalsoup.Browser()
     response = browser._request(form)
 
-    # Check that only "files" includes a "pic" keyword in the response
+    if enctype not in valid_enctypes_file_submit.keys():
+        enctype = default_enctype
+    assert enctype in response.request.headers["Content-Type"]
+
+    resp = response.json()
+    assert resp["form"]["in"] == "test"
+
     found = False
-    for key, value in response.json().items():
-        if key == "files":
-            if "pic" in value:
-                if value["pic"].encode() == expected_content:
-                    # If pic is found twice, an error will occur
-                    assert not found
-                    found = True
-        # One would expect to find "pic" in files, but as of writing,
-        # httpbin puts it in form when the filename is empty:
-        elif key == "form":
-            if "pic" in value:
-                if value["pic"].encode() == expected_content:
-                    assert not found
-                    found = True
-                    assert b"filename=\"\"" in response.request.body
-        else:
-            assert (value is None) or ("pic" not in value)
+    foundIn = None
 
-    assert found
-    assert "multipart/form-data" in response.request.headers["Content-Type"]
+    for key, value in resp.items():
+        if value:
+            if "pic" in value:
+                content = value["pic"].encode()
+                assert not found
+                assert key in ("files", "form")
+                found = True
+                foundIn = key
+            if key == "files" and not valid_enctypes_file_submit[enctype]:
+                assert not value
 
-    # In case we created a file for upload, we need to close & delete it
-    if set_value:
-        os.remove(pic_path)
+    assert found == bool(file_field)
+    if file_field:
+        assert expected_content[(valid_enctypes_file_submit[enctype],
+                                submit_file)] == (content, foundIn)
+
+        if valid_enctypes_file_submit[enctype]:
+            if(submit_file):
+                assert ("filename=\"" + pic_path + "\""
+                        ).encode() in response.request.body
+            else:
+                assert b"filename=\"\"" in response.request.body
+
+
+# delete the temporary file
+# os.remove(pic_path)
+# line commented because it is executed multiple times
+# i don't know how to solve it
 
 
 def test__request_select_none(httpbin):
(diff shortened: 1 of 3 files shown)


hemberger:
Thinking about this more, I'd almost prefer to abide by whatever Requests does, rather than introduce a relatively complicated hack around one of our dependencies. Is there a practical use scenario where we really care about forcing the enctype to be different than the behavior in Requests? (And if yes, would it be better to convince the Requests maintainers that this is true?)

MartinDlry:
A practical use scenario could be when a user testes his website with Mechanical Soup. In one of these cases ( form with `enctype="application/x-www-form-urlencoded"` and file inputs), submitting files will work in Mechanical Soup, but in a real browser it won't work, so Mechanical Soup would hide an error in the website code.

As far as I know, the other case (using `application/x-www-form-urlencoded` when `multipart/data-form` specified) won't cause issues but as we don't know what happens on the server, there can possibly exist some server-side data processing that could cause issues with wrong enctype.

The main reason for respecting enctype is to get closer to a real browser behavior, but as Requests doesn't necessarily share this goal, we are not sure that requests maintainers will care. 
However, the feature that we want could be useful for other softwares and users that use Requests, so this feature request make sense.

I've rewritten the tests following @moy advice, so let me know if you want me to submit my code again.

moy:
A use-case was already given my @jsm28 in #242 ("for testing web applications which behave differently when you submit with the wrong enctype", which I understand essentially as "which may behave differently but should not"). And even if, in practice, the enctype shouldn't really matter, mimicking the behavior of real browsers is usually a good idea IMHO.

Certainly, I'd prefer if requests had an option to explicitly specify the enctype (it does have an option to override the Content-Type field, but it doesn't impact how the payload is actually encoded). I'm all for submitting an issue and/or PR there. However, until requests gets such feature and a release containing it is widely deployed, having a hack in MechanicalSoup does the trick without requiring more dependency. @MartinDlry: willing to discuss with the requests people about it?

@MartinDlry: in any case, do re-submit your code (don't open a new PR, just force-push into this one). Even if we end up rejecting the PR, it's always good to have the latest version of the code available in case we change our minds eventually.

moy:
BTW, the commit also lacks a proper commit message, explaining _why_ this is a good change (i.e. summary of the discussion above) and why the way to do it is the best possible.

tests/test_stateful_browser.py
```diff
@@ -323,8 +323,10 @@ def make_file(content):
                     ("first file content", "second file content"))
 
     # The form doesn't have a type=file field, but the target action
-    # does show it => add the fields ourselves.
+    # does show it => add the fields ourselves, and add enctype too
```
moy:
```suggestion
    # does show it => add the fields ourselves, and add enctype too.
```

tests/test_browser.py
```diff
@@ -112,6 +112,56 @@ def test__request_file(httpbin):
...
+@pytest.mark.parametrize("enctype", [
+  pytest.param("multipart/form-data"),
+  pytest.param("application/x-www-form-urlencoded"),
+  pytest.param("Invalid encytpe")
+])
+def test_specified_enctype(httpbin, enctype, try_submit_file):
+    # test if enctype is respected when specified
+    # and if files are processed correctly
+    form_html = ("""
+    <form method="post" action="{}/post" enctype=\"""" + enctype + """\">
```
moy:
```suggestion
    <form method="post" action="{}/post" enctype="{}">
```

tests/test_browser.py
```diff
@@ -112,6 +112,56 @@ def test__request_file(httpbin):
...
+      <input name="in" value="test" />
+      <input name="fileinput" type="file" />
+    </form>
+    """).format(httpbin.url)
```
moy:
```suggestion
    """).format(httpbin.url, enctype)
```

tests/test_browser.py
```diff
@@ -112,6 +112,56 @@ def test__request_file(httpbin):
...
+    if enctype not in valid_enctypes_file_submit.keys():
+        enctype = default_enctype
+
+    assert enctype in response.request.headers["Content-Type"]
+
+    # response must contain files if and only if the enctype support
+    # sending files and if we submitted files
+    assert bool(resp["files"]) == (valid_enctypes_file_submit[enctype]
+                                   and try_submit_file)
+
```
moy:
```suggestion
if valid_enctypes_file_submit[enctype] and try_submit_file:
    assert resp["files"] == ...
```

MartinDlry:
I've removed a test that is included in the new one, and written a longer commit message.

tests/test_browser.py
```diff
@@ -81,35 +81,65 @@ def test__request(httpbin):
...
+])
+@pytest.mark.parametrize("enctype", [
+  pytest.param("multipart/form-data"),
+  pytest.param("application/x-www-form-urlencoded"),
+  pytest.param("Invalid encytpe")
+])
+def test_enctype_and_file_submit(httpbin, enctype, try_submit_file):
+    # test if enctype is respected when specified
+    # and if files are processed correctly
+    form_html = ("""
```
moy: Why this `(`? `"""` already protect you from newlines.

tests/test_browser.py
```diff
@@ -81,35 +81,65 @@ def test__request(httpbin):
...
+    <form method="post" action="{}/post" enctype="{}">
+      <input name="in" value="test" />
+      <input name="fileinput" type="file" />
     </form>
-    """.format(httpbin.url)
+    """).format(httpbin.url, enctype)
     form = BeautifulSoup(form_html, "lxml").form
 
-    # create a temporary file for testing file upload
```
moy: I'd keep the comment here, but maybe the code is self-explicit enough.

tests/test_browser.py
```diff
@@ -81,35 +81,65 @@ def test__request(httpbin):
...
+        # test if only "files" includes "fileinput" keyword
+        for key, value in resp.items():
+            if key == "files":
+                assert value["fileinput"] == ":-)"
+                found = True
+            else:
+                assert (value is None) or ("fileinput" not in value)
+
+        assert found
+
```
moy: Useless newline.

tests/test_browser.py
```diff
@@ -81,35 +81,65 @@ def test__request(httpbin):
...
+    assert resp["form"]["in"] == "test"
+
+    if enctype not in valid_enctypes_file_submit.keys():
+        enctype = default_enctype
+
+    assert enctype in response.request.headers["Content-Type"]
+
+    # response must contain files if and only if the enctype support
+    # sending files and if we submitted files
+    if (valid_enctypes_file_submit[enctype] and try_submit_file):
```
moy: Useless parenthesis.

moy:
BTW, at some point you should check with @senabIsShort how to merge this one and #250, as they will probably conflict in the tests part. One option is to wait until the first is merged and fix the conflicts after that.

MartinDlry:
These errors were remaining for previous changes I did.

tests/test_browser.py
moy: I still find this newline useless and disturbing (I'd expect the `if` to be over when I see this newline, and I'm surprised by the `else:` right below).

moy:
OK, now that #270 is merged, this branch has conflicts with master. @MartinDlry, can you rebase on top of master and force-push?

Thanks,

moy:
FYI, this PR was written as part of a school project, which is now over.

@MartinDlry, @senabIsShort: can you push the current state of your conflict resolution? You may want to open a new PR and keep this one for the record (it may be helpful to keep an eye on this version while working on the conflict resolution).

Are you interested in finishing the work, or do you want us to take over?

Thanks,

MartinDlry:
I've tried to merge changes. I have an issue with the temporary file, I've put the file creating before the function and file deleting after but the file is deleted before the end of the tests.

This is not finished, i need to add the case with no file input field to achieve 100% code coverage.

tests/test_browser.py
```diff
@@ -81,55 +81,88 @@ def test__request(httpbin):
...
+                              "application/x-www-form-urlencoded": False
+                              }
+
+default_enctype = "application/x-www-form-urlencoded"
+
+# create a temporary file for testing file upload
+file_content = b":-)"
+pic_filedescriptor, pic_path = tempfile.mkstemp()
+os.write(pic_filedescriptor, file_content)
+os.close(pic_filedescriptor)
```
moy: This should really go within the test function. What's the issue when you do so?

tests/test_stateful_browser.py
```diff
@@ -339,8 +339,10 @@ def make_file(content):
                     ("first file content", "second file content"))
 
     # The form doesn't have a type=file field, but the target action
-    # does show it => add the fields ourselves.
+    # does show it => add the fields ourselves, and add enctype too.
     browser.select_form()
+    browser._StatefulBrowser__state.form.form[
+      "enctype"] = "multipart/form-data"
```
moy: The form works within a browser, so it should work so you shouldn't need this.

tests/test_browser.py
MartinDlry: My variable expected_content is defined before the function and there is a case where the file path is needed. The other way to do would be to redefine this entry in the function after creating the temporary file.

tests/test_stateful_browser.py
MartinDlry: The httpbin form is not in multipart/form-data, so you won't be able to upload without adding it, even in a browser

moy: Oops, indeed. The form doesn't have a file entry, but this piece of code is about adding it, you're right.

MartinDlry:
I've found a workaround to put the temporary file creation in the function.

moy:
This PR is not far from being OK. I'm taking over in #290 and closing this one. Thanks @MartinDlry for your work on this, and feel free to comment on what I'm doing in #290.

PR #284, opened 2019-05-23
Fix typos in documentation

tbm:


CONTRIBUTING.rst
```diff
@@ -88,7 +88,7 @@ to view it properly. The file can then be opened in e.g. Inkscape.
 Release Checklist
 -----------------
 
-Releases can be done only by people with sufficient priviledges on
+Releases can be done only by people with sufficient privileges on
```
moy: Fyi : not as wrong as it could be ;-) (https://en.m.wiktionary.org/wiki/priviledge : obsolete form of privilege), but right.

tbm: interesting.

PR #286, opened 2019-06-13
Add a mechanize-like fucntion to parse form inputs

dmknght:
It is like `for form in browser.forms()` in mechanize. It is super useful for automation multiple website, combine with regex and conditions

### DIFF
--- a/mechanicalsoup/form.py
+++ b/mechanicalsoup/form.py
@@ -381,6 +381,23 @@ class Form(object):
                 if subtag.string:
                     subtag.string = subtag.string.strip()
             print(input_copy)
+    def get_info(self):
+    """Get a summary of the form with classic mechanize style
+    """
+    info = "<%s[%s]>\n" %(self.form.get('method'), self.form.get('action'))
+    for input in self.form.find_all(("input", "textarea", "select", "button")):
+        fType = input.attrs['type']
+        try:
+            fValue = input.attrs['value']
+        except:
+            fValue = None
+        try:
+            fID = input.attrs['id']
+        except:
+            fID = input.attrs['name']
+        info += "%s(%s)=\'%s\'\n" %(fType, fID, fValue)
+
+        return info
 
     def find_by_type(self, tag_name, type_attr, attrs):
         attrs_dict = attrs.copy()


hemberger:
Hi, and thanks for your pull request! I'd like to follow up with you so that I understand what your usage is. The function you implemented (`get_info`) looks a lot like `print_summary`, except that it returns the string instead of printing it. I'm not sure I understand how that is useful for automation, though -- I would have thought that the `Form` instance itself would be much easier to work with (since it contains all the information about the form in a structured soup object).

Could you share an example of how you use this function? Thanks!

hemberger:
This pull request **introduces 1 alert** when merging e29cea835f5ce9542782fa800a33d6398bf54386 into 0b04e8b9aed3a3e5160ac89081e2c270598c2f09 - [view on LGTM.com](https://lgtm.com/projects/g/MechanicalSoup/MechanicalSoup/rev/pr-ad9227703d212cb04a99fa907342262a790dbdf6)

**new alerts:**

* 1 for Syntax error

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

*Comment posted by [LGTM.com](https://lgtm.com)*

dmknght:
Hi! For me, it is very useful for working with form automatically. For example: I can get all form fields (instead of just print it) and use regex to parse text field of the form, so I don't have to define textField's value (it works dynamically):
```
textField = re"text\((*.)\)
```
then submit value automatically
```
browser[textField] = value
browser.submit_selected()
```
But you are right about it is totally like print_summary except data return. I think I'll change it a little bit and make something like `browser.forms()` in old mechanize.
P/s: my project with mechanize, auto brute force login form for websites, user don't have to define username and password field name `https://github.com/dmknght/BruteforceHTTP`

moy:
This looks to me like a very fragile way to do what you want. Searching the text may match anything, not just the tags names and attributes. BeautifulSoup was really meant to do this kinds of query, and doing them with BeautifulSoup on the parsed HTML is a much more robust and powerful way to do this.

As you noticed, there are issues with your code (LGTM warning about Python 2 portability and broken CI), but before fixing this I'd rather make sure the code is actually useful and not a duplicate of print_summary().

dmknght:
Hi! Thanks for your information. I've removed this function from `form.py` and created `forms` method in StatefulBrowser.py (this time it is more like old mechanize). I'll test my code carefully.

dmknght:
Close: duplicate function and python syntax

PR #290, opened 2019-07-16
Add "enctype respecting" feature and tests

### LINKED ISSUE #242: Respect enctype specified in forms

moy:
Fixes #242.

Make MechanicalSoup behave as a real browser:

- only "multipart/form-data" enctype allow file sending

- forms without file inputs can now be sent as "multipart/form-data"
  if specified.

- "application/x-www-form-urlencoded" is the default when enctype is wrong
  or not specified

The added code uses a hack because of a lack in Requests
functionalities: Requests doesn't support forcing enctype yet, and
doesn't allow to submit using "multipart/data-form" enctype without
submitting files. So the best way we found to bypass this is to pass
an modified dict as files to Requests. This code could change in the
future if Requests implements a "force enctype" feature.

### DIFF
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -155,6 +155,8 @@ class Browser(object):
         # Requests also retains order when encoding form data in 2-tuple lists.
         data = [(k, v) for k, v in data.items()]
 
+        multipart = form.get("enctype", "") == "multipart/form-data"
+
         # Process form tags in the order that they appear on the page,
         # skipping those tags that do not have a name-attribute.
         selector = ",".join("{}[name]".format(i) for i in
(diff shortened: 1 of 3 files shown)


moy:
OK, I think the code is now ready to be merged. There are multiple features in one (see the bullet list in the commit message), but they are too intricate to be easily split into separate commits. I still split the tests major modification and the actual feature, so that the second commit documents what changes in the tests.

moy:
> Is this change worth the complexity that it adds?

Note that most of the complexity is in tests, and testing all cases "with file input, without / possible enctype" is a good thing IMHO regardless of the actual behavior of MechanicalSoup.

> I am happy with MechanicalSoup wrapping Requests functionality rather than mimicking what browsers do at any cost

There are several things in this PR:

1. only "multipart/form-data" enctype allow file sending

2. forms without file inputs can now be sent as "multipart/form-data"
    if specified.

3. "application/x-www-form-urlencoded" is the default when enctype is wrong
    or not specified

Actually, 1. and 3. are relevant features for people using MechanicalSoup to test their website, as they will miss obvious bugs otherwise (missing `enctype` in `<form ...>`). Honestly, I have no idea why things are done this way, if I were inventing the web today I'd switch to `multipart/form-data` whenever there's a file input, but that's not what browsers do.

Item 2. is more a detail, I can't think of any real-life use-case for it, but it shouldn't harm. The [implementation is a bit hacky](https://github.com/MechanicalSoup/MechanicalSoup/pull/290/files#diff-ae060c6024d83a9454f5bfb9144bdef2R233-R241) and does rely on a requests implementation detail, but it's a very local hack, easy enough to remove if it ever becomes a problem. Since it's not a feature many users are likely to rely on, it's also not a huge problem if we introduce this feature and remove it afterwards if requests totally breaks our code.

So I think the benefits of merging the PR are larger than the cost.

PR #293, opened 2019-08-14
Properties

moy:
This superseeds #185 (which I'm keeping untouched for the record).

Essentially:

* No more getters in the recommended interface, just properties.

* Getters are kept with their old names, only for backward compatibility

* Properties are now used everywhere (examples, code, tests)

Each commit passes tests. PR best reviewed commit-by-commit, as it contains both small manual changes with batch search-and-replace.

### DIFF
--- a/docs/ChangeLog.rst
+++ b/docs/ChangeLog.rst
@@ -8,6 +8,18 @@ Version 1.0 (in development)
 Main changes:
 -------------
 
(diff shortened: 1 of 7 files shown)


moy:
Codacy raises a warning, but it's a false-positive (it didn't complain about `browser.get_selected_form()` and does for `browser.form`). I've marked it as such in Codacy, but the failure seems to still appear here.

PR #294, opened 2019-08-14
Properties

moy:
This superseeds #185 (which I'm keeping untouched for the record), and #293 (which I closed by mistake, messing up with a `git push` command, sorry).

Essentially:

* No more getters in the recommended interface, just properties.

* Getters are kept with their old names, only for backward compatibility

* Properties are now used everywhere (examples, code, tests)

Each commit passes tests. PR best reviewed commit-by-commit, as it contains both small manual changes with batch search-and-replace.

### DIFF
--- a/docs/ChangeLog.rst
+++ b/docs/ChangeLog.rst
@@ -11,6 +11,23 @@ Version 0.12
 Main changes:
 -------------
 
+* ``StatefulBrowser`` methods ``get_current_page``,
+  ``get_current_form`` and ``get_url`` have been deprecated in favor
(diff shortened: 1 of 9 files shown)


moy:
If anyone needs to write code different when a form is selected or not, I agree your `has_selected_form` looks nicer. But to me, the caller of MechanicalSoup is the one who called `select_form()`, and that code knows whether a form is selected or not, at least in 99.9% of cases. The exception is just here for defensive programming and to provide a nice error message.

IOW, why not, but I'm not sure many people would actually use it.

hemberger:
I'm fine with taking the less defensive "add it if requested" stance. :)

PR #300, opened 2019-09-17
Update a few missed instances of deprecated methods

hemberger:
This patch addresses a few omissions from #294.


tests/test_stateful_browser.py
```diff
@@ -265,7 +265,7 @@ def test_new_control(httpbin):
     browser.new_control("text", "comments", "This is an override comment")
     fake_select = BeautifulSoup("", "html.parser").new_tag('select')
     fake_select["name"] = "foo"
-    browser.get_current_form().form.append(fake_select)
```
moy: Ah, it wasn't there when I wrote my PR, that's why I missed it. Obviously correct.

PR #303, opened 2019-11-28
change browser.page by browser.get_current_page()

jbarotin:
browser.page is not working

### DIFF
--- a/examples/example.py
+++ b/examples/example.py
@@ -30,7 +30,7 @@ resp = browser.submit_selected()
 # browser.launch_browser()
 
 # verify we are now logged in
-page = browser.page
+page = browser.get_current_page()
 messages = page.find("div", class_="flash-messages")
 if messages:
     print(messages.text)


moy:
`browser.page` is working, but it's a recent feature, unreleased for now.

@hemberger : shall we release 1.0, BTW?

jbarotin:
As you, I tried the example with the version installed though pip (MechanicalSoup==0.12.0), and it's not working with ".page".  

After reading your documentation (because the ".page" not working), I changed to ".get_current_page()" and then get the example to work...

moy:
By definition, 0.12 is not an unreleased version ;-)

hemberger:
@moy I have no problem with 1.0 being released. The only question I have is whether or not we want to refer to the old methods as "deprecated" in the changelog (which may be stressful to people migrating), or if we want to refer to the new interface as a convenience that *may* one day replace the existing methods.

PR #304, opened 2019-12-11
Parse response only when Content-Type is not set

vvatsava:
Current implementation parses response.text both when content-type is not text/html or empty
This leads parsing data when content-type is binary data such as application/x-gzip.
Parse only when content-type not passed.

### DIFF
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -65,8 +65,9 @@ class Browser(object):
     @staticmethod
     def add_soup(response, soup_config):
         """Attaches a soup object to a requests response."""
-        if ("text/html" in response.headers.get("Content-Type", "") or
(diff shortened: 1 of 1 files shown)


moy:
You can run flake8 locally to avoid going through travis each time. autopep8 is your friend, too if needed ;-).

moy:
@vvatsava: do you have anything to add following up to @hemberger's comment?

There's still a flake8 warning, but that's easy to fix (at worse, we can do that while merging).

Thanks,

hemberger:
Without any follow-up, I'm inclined to prefer the current implementation (where we parse the response so long as we're told it's HTML _or_ it looks like it's HTML, regardless of what we're told). Please feel free to reopen this PR at any time if you'd like to continue this discussion. Thanks again!

Issue #310, opened 2020-01-16
drop support for Python 2


moy:
Python 2's maintenance ends in 2020, and that happens to be now (there should be [one last release of Python 2 in april](https://adtmag.com/articles/2020/01/09/python-2-end-of-life.aspx), and then nothing).

We may start thinking about terminating support for Python 2 in MechanicalSoup.

I'd like to (but I can live without it) have MechanicalSoup 1.0 support Python 2, so that an easy way for people who really need it with MechanicalSoup corresponds to a version with a round number. Anyway, we can start writing a PR to drop support and remove any code needed only for Python 2. A quick grep for `Python 2` and `__future__` shows that there should be only very marginal change (i.e. only lines to remove, and `.travis-ci.yml` to adapt).

hemberger:
I'm perfectly happy to have Python 2 support in MechanicalSoup 1.0. It will likely stay relevant for a while even after it is officially unsupported.

hugovk:
Here's the pip installs for MechanicalSoup from PyPI for December 2019:

| category | percent | downloads |
|----------|--------:|----------:|
| 3.7      |  42.14% |    10,777 |
| 3.6      |  33.04% |     8,449 |
| 2.7      |  12.32% |     3,150 |
| 3.8      |   6.51% |     1,664 |
| null     |   3.12% |       799 |
| 3.5      |   2.71% |       694 |
| 3.4      |   0.11% |        29 |
| 3.9      |   0.04% |        11 |
| Total    |         |    25,573 |

Source: `pip install -U pypistats && pypistats python_minor MechanicalSoup --last-month`

12% for Python 2.7 is pretty low, especially compared to some other projects:

* https://dev.to/hugovk/python-version-share-over-time-6-1jb8

moy:
What I had in mind was also: https://python3statement.org/

In short, clearly there are still Python 2 users out there, but at some point library maintainers should start pushing for Python 3 a bit harder, and stop spending effort maintaining support for Python 2. In our case, the effort to support Python 2 is very small so it's not a real argument, but we may help (at our very small scale) spreading the message "guys, if you didn't migrate to Python 3 yet, now is the time to do so".

And also, it's a very easy task to write the PR, a good warmup task for someone willing to contribute to MechanicalSoup ;-). Merging the PR can be deferred as much as needed after that.

hugovk:
[...]

PR #311, opened 2020-01-19
Add python_requires to help pip

hugovk:
If there's one last release before dropping Python 2 (https://github.com/MechanicalSoup/MechanicalSoup/issues/310), this will help pip install the right version of MechanicalSoup for their running Python, and ease migration.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,6 +52,8 @@ setup(
 
     license=about['__license__'],
 
+    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
+
     classifiers=[
         'License :: OSI Approved :: MIT License',
 


hemberger:
In case it isn't clear, I'm happy to defer to those who are more knowledgeable about the packaging infrastructure, but I'd be grateful for some references or details about why things are done the way they are done. The official docs don't really give insight into the "conventional wisdom".

Thanks again for the PR's!

hugovk:
It's possible to add an upper limit too:

```python
    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, <3.9',
```

That would of course need a new release as soon as you want to allow installation on 3.9, including people wanting to test the alpha right now (but they can use `--ignore-requires-python`).

---

What sort of non-pip installations could there be?

This metadata (`Requires-Python`) is part of a standard ([PEP 345](https://www.python.org/dev/peps/pep-0345/)), so is available for other installation tools to use it. So I suppose tools like conda could use it too, but I'm not very familiar with conda.

Other tools should ignore the metadata if they don't support it, and behave as they do now.

hemberger:
Doing a little more legwork (sorry for the spam), is [this](https://packaging.python.org/guides/distributing-packages-using-setuptools/#classifiers) the relevant documentation snippet?
> Although the list of classifiers is often used to declare what Python versions a project supports, this information is only used for searching & browsing projects on PyPI, not for installing projects. To actually restrict what Python versions a project can be installed on, use the python_requires argument.

So this is presumably why we need to specify both? And the Classifiers are also used to construct the supported versions badge, I suppose. Do we not specify an upper bound on the version in `python_requires` to avoid (likely) unnecessarily restricting users of the latest interpreter, even if we haven't yet tested our compliance?

hugovk:
You're welcome! 

There's a related discussion here about whether to set an upper limit of `<4`, and they suggest not to, because there's no indication Python 4 would be a breaking change. 

https://discuss.python.org/t/use-of-less-than-next-major-version-e-g-4-in-python-requires-setup-py/1066

3.9 has less reason to be a breaking change (beyond normal deprecation removals).

An upper limit would definitely be useful for a package which only supports Python 2 and not 3.

---

Yep, that docs snippet is relevant. Yep, Trove classifiers are used for badges and also on the PyPI website, like https://pypi.org/project/MechanicalSoup/#data. 

Thanks!

PR #315, opened 2020-02-25
Update expl_duck_duck_go.py

dillonko:
This will fix the issue of finding the search field until it gets changed again.

### DIFF
--- a/examples/expl_duck_duck_go.py
+++ b/examples/expl_duck_duck_go.py
@@ -12,7 +12,7 @@ browser = mechanicalsoup.StatefulBrowser()
 browser.open("https://duckduckgo.com/")
 
 # Fill-in the search form
-browser.select_form('#search_form_homepage')
+browser.select_form('#search_form_input_homepage')
 browser["q"] = "MechanicalSoup"
 browser.submit_selected()
 


dillonko:
Thanks will do!

hemberger:
Unfortunately, while this PR is much appreciated, I don't think it is correct.

The Duck Duck Go form looks like this:
```html

<form id="search_form_homepage" class="search  search--home  js-search-form" name="x" method="POST" action="/html">
  <input id="search_form_input_homepage" class="search__input  js-search-input" type="text" autocomplete="off" name="q" tabindex="1" value="">
  <input id="search_button_homepage" class="search__button  js-search-button" type="submit" tabindex="2" value="S" />
  <input id="search_form_input_clear" class="search__clear  empty  js-search-clear" type="button" tabindex="3" value="X" />
  <div id="search_elements_hidden" class="search__hidden  js-search-hidden"></div>
</form>
```
Two things:
1. `#search_form_homepage` was still the correct selector for the form. `#search_form_input_homepage` is the selector for the `<input>` element. The author of this PR should have gotten a warning saying 
    > Constructed a Form from a 'input' instead of a 'form' element. This may be an error in a future version of MechanicalSoup.

    but did not, because apparently `PendingDeprecationWarning` warnings are suppressed by default! I will submit a PR fixing this shortly (my apologies to @dillonko!).

2. With this PR reverted, the body of the response is:
    ```html
    <html><body><p>If this error persists, please let us know: error-lite@duckduckgo.com</p></body></html>
    ```
   @moy had already identified this in 89ec62a5e741614546d078408129d488dae40e0b (suggesting an anti-bot filter). Maybe we should just remove this example?

PR #318, opened 2020-02-29
Revert "Update expl_duck_duck_go.py"

hemberger:
Reverts MechanicalSoup/MechanicalSoup#315

### DIFF
--- a/examples/expl_duck_duck_go.py
+++ b/examples/expl_duck_duck_go.py
@@ -12,7 +12,7 @@ browser = mechanicalsoup.StatefulBrowser()
 browser.open("https://duckduckgo.com/")
 
 # Fill-in the search form
-browser.select_form('#search_form_input_homepage')
+browser.select_form('#search_form_homepage')
 browser["q"] = "MechanicalSoup"
 browser.submit_selected()
 


hemberger:
Can do -- I'll summarize https://github.com/MechanicalSoup/MechanicalSoup/pull/315#issuecomment-591096944 in the commit message. But I also posited that we should probably just remove this example entirely, since it will likely never work again.

moy:
Removing the example instead of un-breaking it partly is fine with me. Finding another search engine that works with us would be even better.

dillonko:
Hey all,

I understand that the change could possibly break again. I can write a solution that always grabs the search box ID and queries that. I am a big fan of your tool which is why I wanted to contribute to your examples.

I also understand that because it is a webpage that you are interacting with it is always subject to change. I want to help out more and am new to the PR space outside of work how can I better help. What would be a better search engine to use? Thank you all for your considerations.

Dillon KO :)

moy:
> I understand that the change could possibly break again.

The problem is not that it may break again, but that it is already broken. I merged your PR without testing, but actually, with the old version I get:

```
$  python3 ./expl_duck_duck_go.py
$ 
```

Looking at the results page, the HTML is:

```
<html><body><p>If this error persists, please let us know: error-lite@duckduckgo.com</p></body></html>
```

which indicates an error server-side (hence the absence of link in the result), rather likely an anti-bot measure.

With your patch, I get:

```
$ python3 ./expl_duck_duck_go.py
/home/mmoy/dev/MechanicalSoup/mechanicalsoup/form.py:41: FutureWarning: Constructed a Form from a 'input' instead of a 'form' element. This may be an error in a future version of MechanicalSoup.
  .format(form.name), FutureWarning)
  File "./expl_duck_duck_go.py", line 16, in <module>
    browser["q"] = "MechanicalSoup"
  File "/home/mmoy/dev/MechanicalSoup/mechanicalsoup/stateful_browser.py", line 120, in __setitem__
    self.form[name] = value
  File "/home/mmoy/dev/MechanicalSoup/mechanicalsoup/form.py", line 237, in __setitem__
    return self.set(name, value)
  File "/home/mmoy/dev/MechanicalSoup/mechanicalsoup/form.py", line 278, in set
    raise LinkNotFoundError("No valid element named " + name)
mechanicalsoup.utils.LinkNotFoundError: No valid element named q
$ 
```

So it crashes earlier in the script and doesn't event reach the `submit_form` statement.

Does the script work for you? If so, can you post an example input-output you get with it on your machine?

moy:
@dillonko do you have a working `expl_duck_duck_go.py`? If so, any help would be appreciated to make it work for us too. If not, then we should just delete this file (better than merging this PR IMHO).

dillonko:
It worked for me the last time I will test it again today.

hemberger:
@moy Agreed. I'd rather delete the example altogether than merge this PR (assuming it cannot be fixed).

hemberger:
Superseded by #325

PR #319, opened 2020-03-02
Example use of MechanicalSoup using Qwant

moy:
Our DuckDuckGo example doesn't work anymore, probably due to an
anti-bot feature on DuckDuckGo. This introduces and promotes in
README.rst an example that actually works.

Unless we find a way to repair the DuckDuckGo example, we probably want to remove examples/expl_duck_duck_go.py completely instead of #318.

Cc: @dillonko

### DIFF
--- a/examples/expl_qwant.py
+++ b/examples/expl_qwant.py
@@ -1,0 +1,26 @@
+"""Example usage of MechanicalSoup to get the results from the Qwant
+search engine.
+"""
+
+import re
+import mechanicalsoup
+import urllib.parse
+
+# Connect to duckduckgo
+browser = mechanicalsoup.StatefulBrowser(user_agent='MechanicalSoup')
+browser.open("https://lite.qwant.com/")
+
+# Fill-in the search form
+browser.select_form('#search-form')
+browser["q"] = "MechanicalSoup"
+browser.submit_selected()
+
+# Display the results
+for link in browser.page.select('.result a'):
+    # Qwant shows redirection links, not the actual URL, so extract
+    # the actual URL from the redirect link:
+    href = link.attrs['href']
+    m = re.match(r"^/redirect/[^/]*/(.*)\?.*$", href)
+    if m:
+        href = urllib.parse.unquote(m.group(1))
+    print(link.text, '->', href)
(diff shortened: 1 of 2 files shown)


dillonko:
Thanks I will share the working implementation and work on capturing the
form data so that it doesn't run into this issue again.

I find this all really interesting.  Thanks for reaching out.

Best,
Dillon KO

examples/expl_qwant.py
```diff
@@ -0,0 +1,26 @@
...
+browser.open("https://lite.qwant.com/")
+
+# Fill-in the search form
+browser.select_form('#search-form')
+browser["q"] = "MechanicalSoup"
+browser.submit_selected()
+
+# Display the results
+for link in browser.page.select('.result a'):
+    # Qwant shows redirection links, not the actual URL, so extract
```
hemberger: For better or worse, this also shows the ads at the top of the list of results. It looks like the real results _all_ match the regexp, so we could only print those (and omit the ones that are direct links). However, the redirect link also has a bunch of query string params that only apply to the qwant page.

Alternatively, it may be more straightforward to get the url from:
```html
<div class="url">https://<b>mechanicalsoup</b>.readthedocs.io/en/stable</div>
```
(and just strip the html tags using `.text`).

moy: > For better or worse, this also shows the ads at the top of the list of results. It looks like the real results all match the regexp, so we could only print those

I saw that, but I chose to keep the logic as simple as it could be (just list all links in the page), and possibly as robust as it could be too (in case the redirect scheme changes in the future). If you prefer the version without ads, it's as simple as this:

```
diff --git a/examples/expl_qwant.py b/examples/expl_qwant.py
index c10e4ab..af27f7d 100644
--- a/examples/expl_qwant.py
+++ b/examples/expl_qwant.py
@@ -23,4 +23,4 @@ for link in browser.page.select('.result a'):
     m = re.match(r"^/redirect/[^/]*/(.*)\?.*$", href)
     if m:
         href = urllib.parse.unquote(m.group(1))
-    print(link.text, '->', href)
+        print(link.text, '->', href)
```

> However, the redirect link also has a bunch of query string params that only apply to the qwant page.

Fixed. Fortunately, the `?...` part appears unquoted in the redirect URL, hence it's straightforward to filter.

> Alternatively, it may be more straightforward to get the url from: `<div class="url">...`

Not really: long URLs are cut in this `<div>` (i.e. end of URL replaced with `...`). This field is really meant to be displayed to the user, not machine-consumed.

Issue #322, opened 2020-04-28
get input value


ism:
Hi there everyone!

I am trying to get input form value of selected form. I came up with this

```
b = mechanicalsoup.StatefulBrowser()
...
firstname = b.get_current_form().form.find('input', {'name': 'firstname'})['value']

```
Is there any better way to do this?
As a suggestion, maybe we can get values from forms as easy as setting them?
In my case it would look like that:

`firstname = b['firstname']`

moy:
Indeed, we do have `__setitem__` methods in `StatefulBrowser` and `Form` to allow the `[...] = ...` syntax, but no `__getitem__` to read from a field. It should be rather straightforward to implement. Not sure I'll find time to do it soon, but this looks like a nice first patch to make if you're interested in contributing ;-).

ism:
I am interested :-)

I have added support for 'input' type. Should i also add support for checkbox, radio, textarea and select? I am asking because, for example, in case of checkbox it's not clear what type to return - True, 'checked', None ?

moy:
Best is to submit a minimalist and uncontroversial PR, and then we'll see.

I don't have the whole code/behavior in mind, but the goal should be to be as symmetrical as possible to the `set_*` methods in `Form`.

dillonko:
I belive you can do something like this:

```python
b = mechanicalsoup.StatefulBrowser()
...
firstname = b.get_current_form().form.find('input', {'name': 'firstname'}).get('value')
```

pcworld:
This appears unsolved, although there's a commit at https://github.com/ism/MechanicalSoup/commit/a71c43abaf68e22bafe3c2290193deb30a72eb62, there doesn't seem to be a pull request for it.

PR #323, opened 2020-04-29
Add ability to get form input fields

ism:
Issue #322 

added ability to get input fields like
`a = browser['a']`

and ability to test if given input field is present in form fields
`if 'field' in browser:
...
`

My first PR, i might have missed something...

### DIFF
--- a/mechanicalsoup/form.py
+++ b/mechanicalsoup/form.py
@@ -231,7 +231,7 @@ class Form(object):
                 option.attrs["selected"] = "selected"
 
     def __setitem__(self, name, value):
-        """Forwards arguments to :func:`~Form.set`. For example,
+        """Forward arguments to :func:`~Form.set`. For example,
         :code:`form["name"] = "value"` calls :code:`form.set("name", "value")`.
         """
         return self.set(name, value)
@@ -276,6 +276,44 @@ class Form(object):
             self.new_control('text', name, value=value)
             return
         raise LinkNotFoundError("No valid element named " + name)
+
+    def __getitem__(self, name):
+        """Forward arguments to :func:`~Form.get` to get value for a given form field.
+        For example, :code:`name = form["name"]` calls :code:`form.get("name")`
+        """
+        return self.get(name)
+
+    def get(self, name):
+        """Get a ``value`` from input element identified by ``name``.
+        Currently it does support "input" type only.
+
+        Example:
+
+        .. code-block:: python
+            firstname = browser['firstname']
+
+        If the element is not found , raise a :class:`LinkNotFoundError`
+        exception.
+
+"""
+        inp = self.form.find("input", {"name": name})
+        if not inp:
+            raise LinkNotFoundError("No input field named " + name)
+        return inp["value"]
+
+    def __contains__(self, name):
+        """Test if ``name`` input field is in the form.
+        Support "input" type only.
+
+        Example:
+
+        .. code-block:: python
+           if 'name' in form:
+               ...
+
+        """
+        inp = self.form.find("input", {"name": name})
+        return True if inp else False
 
     def new_control(self, type, name, value, **kwargs):
         """Add a new input element to the form.
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -546,5 +546,45 @@ def test_option_without_value(fail, selected, expected_post):
         assert res.status_code == 200 and res.text == 'Success!'
 
 
+one_field_form = '''<form method="post">
+  <input name="life" type="hidden" value="good"/>
+</form>'''
+
+
+def test_get_value_of_input_field():
+    """Test that we can get input field of current form by name"""
+
+    browser, url = setup_mock_browser(text=one_field_form)
+    browser.open(url)
+    browser.select_form()
+
+    assert browser['life'] == 'good'
+    browser['life'] = 'veryGood'
+    assert browser['life'] == 'veryGood'
+
+
+def test_getting_invalid_field_raises_exception():
+    """Test that we can get input field of current form by name"""
+
+    browser, url = setup_mock_browser(text=one_field_form)
+    browser.open(url)
+    browser.select_form()
+
+    with pytest.raises(mechanicalsoup.utils.LinkNotFoundError):
+        _ = browser['unknown_life']
+
+
+def test_if_given_field_is_in_current_form():
+    """Test if given input field is one of form input fields"""
+    browser, url = setup_mock_browser(text=one_field_form)
+    browser.open(url)
+    browser.select_form()
+
+    assert 'life' in browser
+    assert 'unknown_life' not in browser
+    browser.new_control('text', 'unknown_life', 'goodToo')
+    assert 'unknown_life' in browser
+
+
 if __name__ == '__main__':
     pytest.main(sys.argv)
--- a/mechanicalsoup/stateful_browser.py
+++ b/mechanicalsoup/stateful_browser.py
@@ -119,6 +119,18 @@ class StatefulBrowser(Browser):
         """
         self.form[name] = value
 
+    def __getitem__(self, name):
+        """Get input value of a field in currently selected form.
+        See :func:`Form.__getitem__`.
+        """
+        return self.form[name]
+
+    def __contains__(self, name):
+        """Test if specific field is in the form's input fields.
+        See :func:`Form.__contains__`.
+        """
+        return name in self.form
+
     def new_control(self, type, name, value, **kwargs):
         """Call :func:`Form.new_control` on the currently selected form."""
         return self.form.new_control(type, name, value, **kwargs)


mechanicalsoup/form.py
```diff
@@ -277,6 +277,24 @@ def set(self, name, value, force=False):
...
 
+    def __getitem__(self, name):
+        """Forwards arguments to :func:`~Form.get` to get value for a given form field
+        """
+        return self.get(name)
+
+    def get(self, name):
+        i = self.form.find("input", {"name": name})
+        if not i:
+            raise LinkNotFoundError("No input field named " + name)
```
moy, line 288: We've already been using `LinkNotFoundError` for too many things, but I think it's time to stop this proliferation.

Should we add a `ElementNotFoundError` or so?

mechanicalsoup/form.py
```diff
@@ -277,6 +277,24 @@ def set(self, name, value, force=False):
             return
         raise LinkNotFoundError("No valid element named " + name)
 
+    def __getitem__(self, name):
+        """Forwards arguments to :func:`~Form.get` to get value for a given form field
+        """
+        return self.get(name)
+
+    def get(self, name):
```
moy, line 285: This lacks a docstring, which is especially frustrating since `__getitem__` explicitly points to this method.

mechanicalsoup/form.py
```diff
@@ -277,6 +277,24 @@ def set(self, name, value, force=False):
             return
         raise LinkNotFoundError("No valid element named " + name)
 
+    def __getitem__(self, name):
+        """Forwards arguments to :func:`~Form.get` to get value for a given form field
```
moy, line 281: We usually use an imperative tone for docstrings ("dear computer, please forward ..."), hence no 's'.

tests/test_form.py
```diff
@@ -546,5 +546,43 @@ def test_option_without_value(fail, selected, expected_post):
...
+
+
+def test_get_value_of_input_field():
+    """Test that we can get input field of current form by name"""
+
+    browser, url = setup_mock_browser(text=one_field_form)
+    browser.open(url)
+    browser.select_form()
+
+    assert browser['life'] == 'good'
```
moy, line 561: While you're there, checking that assignment assigns to the same place as reading would be nice:

```suggestion
    assert browser['life'] == 'good'
    browser['life'] = 'veryGood'
    assert browser['life'] == 'veryGood'
```

tests/test_form.py
```diff
@@ -546,5 +546,43 @@ def test_option_without_value(fail, selected, expected_post):
...
+
+
+def test_if_given_field_is_in_current_form():
+    """Test if given input field is one of form input fields"""
+    browser, url = setup_mock_browser(text=one_field_form)
+    browser.open(url)
+    browser.select_form()
+
+    assert 'life' in browser
+    assert 'unknown_life' not in browser
```
moy, line 584: Same as above, I'd also check that modifications made to the form are properly reflected by the operators.
```suggestion
    assert 'unknown_life' not in browser
    browser.new_control('text', 'unknown_life', 'goodToo')
    assert 'unknown_life' in browser
```

?

mechanicalsoup/form.py
ism: I had KeyError in mind, but then found this.
https://github.com/MechanicalSoup/MechanicalSoup/blob/b6307dfea36a57a3a8a97674e72829d3f07b890e/mechanicalsoup/utils.py#L12-L14

So I kept it for consistency. Indeed, its good idea to use ElementNotFound Exception, but again for consistency .set method should use it too?

mechanicalsoup/form.py
ism: Any tips how should i proceed with all these suggestions? git push --force or additional commit?

mechanicalsoup/form.py
moy: We started using `LinkNotFoundError` in places where we shouldn't have. The issue is backward compatibility, which forces us to keep code catching `LinkNotFoundError` working with future versions of the library.

Maybe we should just declare a new `MechanicalSoupError` alias for `LinkNotFoundError`, and use it. We'd keep consistency, backward compatibility, but we wouldn't have distinct exceptions for specific cases.

mechanicalsoup/form.py
moy: In the end, we'll want a clean history (so given the scope of the PR, most likely only one commit). Sometimes it makes sense to start with additional commits during the review, but `push --force` is probably best here.

mechanicalsoup/form.py
ism: This one is not easy to fix, it might require additional discussion. I would use MechanicalSoupError as base exception class, with derived classes like LinkNotFoundError, ElementNotFoundError and InvalidFormMethod. I would prefer to use separate PR for such change, otherwise this one might never be finished. 🙂

moy: I was pointing it out because the longer we wait, the harder it may be to fix later, but you're right that it shouldn't block merging this PR.

hemberger:
I would like to carefully review this, but I will need a little bit more time before I can do that. At first glance, though, this looks great!

mechanicalsoup/form.py
```diff
@@ -277,6 +277,45 @@ def set(self, name, value, force=False):
             return
         raise LinkNotFoundError("No valid element named " + name)
 
+    def __getitem__(self, name):
+        """Forward arguments to :func:`~Form.get` to get value for a given form field.
+        For example, :code:`name = form["name"]` calls :code:`form.get("name")`
+        .
```
hemberger, line 283: Any reason this period is on its own line? Should probably just be on the previous line.

mechanicalsoup/form.py
```diff
@@ -277,6 +277,45 @@ def set(self, name, value, force=False):
...
+        Example:
+
+        .. code-block:: python
+            firstname = browser['firstname']
+
+        If the element is not found , raise a :class:`LinkNotFoundError`
+        exception.
+
+"""
+        i = self.form.find("input", {"name": name})
```
hemberger, line 300: I think it'd be good to avoid single character variables if possible. Even `inp` instead of `i` would be good here.

mechanicalsoup/form.py
```diff
@@ -277,6 +277,45 @@ def set(self, name, value, force=False):
...
+        Support "input" type only.
+
+        Example:
+
+        .. code-block:: python
+           if 'name' in form:
+               ...
+
+        """
+        i = self.form.find("input", {"name": name})
```
hemberger, line 316: I was trying to think if there was any way to avoid duplication of this particular line between `__contains__` and `get`, but it seems likely that any attempt to do so would actually be more verbose. (But again this could be `inp` instead of `i`).

tests/test_form.py
```diff
@@ -546,5 +546,47 @@ def test_option_without_value(fail, selected, expected_post):
...
+    """Test that we can get input field of current form by name"""
+
+    browser, url = setup_mock_browser(text=one_field_form)
+    browser.open(url)
+    browser.select_form()
+
+    assert browser['life'] == 'good'
+    browser['life'] = 'veryGood'
+    assert browser['life'] == 'veryGood'
+    browser.close()
```
hemberger, line 564: We don't need to explicitly specify `close` anymore. This line can be removed.

tests/test_form.py
```diff
@@ -546,5 +546,47 @@ def test_option_without_value(fail, selected, expected_post):
...
+def test_getting_invalid_field_raises_exception():
+    """Test that we can get input field of current form by name"""
+
+    browser, url = setup_mock_browser(text=one_field_form)
+    browser.open(url)
+    browser.select_form()
+
+    with pytest.raises(mechanicalsoup.utils.LinkNotFoundError):
+        _ = browser['unknown_life']
+    browser.close()
```
hemberger, line 576: We don't need to explicitly specify `close` anymore. This line can be removed.

hemberger (approved):
Thanks for the PR! I have a few very minor suggestions for improvements, but since I've been so slow to review this (my sincere apologies for that!), I'm fine with merging it as-is, and we can address these remaining issues separately if we want to.

ism:
Hello guys,
i have updated commit with suggested changes. Now it fails few checks. Any tips how to make it happy?

hemberger:
From Travis:
```
mechanicalsoup/form.py:283:80: E501 line too long (80 > 79 characters)
```
Looks like the first line of the docstring for `form.__getitem__` is 1 character too long.

I'm not seeing where the test coverage loss is from. Maybe try rebasing against `upstream/master` and force push to your branch? (Since there are a number of commits that have been applied to the project since this PR was created.)

Thanks for the updates!

ism:
@hemberger thank you! I managed to pass all checks now :)

mechanicalsoup/form.py
```diff
@@ -277,6 +277,44 @@ def set(self, name, value, force=False):
...
+            firstname = browser['firstname']
+
+        If the element is not found , raise a :class:`LinkNotFoundError`
+        exception.
+
+"""
+        inp = self.form.find("input", {"name": name})
+        if not inp:
+            raise LinkNotFoundError("No input field named " + name)
+        return inp["value"]
```
moy, line 302: It is suspicious that we have such complex logic in `set` and that `get` is so straightforward.

I think your code works only on text fields. For example, for radio buttons you need to look for the option having `checked="checked"`. This is more or less what `Browser.get_request_kwargs` does.

It's OK to deal only with text fields as a limitation, but the code should throw `NotImplementedError` for non-managed types of fields (so that we can later add proper management for them without breaking compatibility).

moy, line 302: It is suspicious that we have such complex logic in `set` and that `get` is so straightforward.

I think your code works only on text fields. For example, for radio buttons you need to look for the option having `checked="checked"`. This is more or less what `Browser.get_request_kwargs` does.

It's OK to deal only with text fields as a limitation, but the code should throw `NotImplementedError` for non-managed types of fields (so that we can later add proper management for them without breaking compatibility).

moy (changes requested):
See comment above, the PR should either properly deal with all kinds of fields (hard) or raise an exception (easy, probably the right way in a first PR, which can be improved later).

mechanicalsoup/form.py
ism: It indeed does, as state in PR message it works only with 'input' type fields

moy:
It's a pity that the PR was closed, it did add an interesting feature. Raising `NotImplementedError` should be done before this reaches `master`, but I could have done this while merging. I can't now, since the branch is deleted.

If you're not used to it, it's normal that a review triggers all kinds of comments and requests for changes. It's how the code gets better, and how programmers get better. There's nothing personal.

PR #324, opened 2020-05-03
Update expl_duck_duck_go.py

dillonko:
After testing a few times and debugging found the solution. Documentation should probably be updated


hemberger:
Hey @dillonko, thanks for the PR!

It turns out that the only change that was necessary is adding a user agent. Since this includes a bunch of other changes, I'll close this in favor of a PR that only changes the user agent. I've attributed the solution to you in the commit message -- thanks so much for the tip and for your continued work to help resolve this issue!

dillonko:
COOL thanks!

PR #326, opened 2020-05-03
Update expl_duck_duck_go.py

dillonko:
This is the proper working example

### DIFF
--- a/examples/expl_duck_duck_go.py
+++ b/examples/expl_duck_duck_go.py
@@ -6,13 +6,17 @@ import mechanicalsoup
 
 # Connect to duckduckgo
 browser = mechanicalsoup.StatefulBrowser(user_agent="MechanicalSoup")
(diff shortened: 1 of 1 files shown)


hemberger:
Hi, and thanks again for the help!

I appreciate your concern for this example, but I don't think these changes are needed. The main substantive change is from `browser.page.select('a.result__a')` to `browser.links()`, but the latter actually slurps in too many hyperlinks (we get a lot of duplicates).

If I'm misunderstanding the problem you're seeing, please don't hesitate to let us know!

dillonko:
Yeah, no problem. It looks like examples only work for python 2. So the doc
strings should probably reflect that.

I think that is what I'm trying to say. The tests you have probably won't
reflect that since I get the feeling that you are using __future__ to test
for python three implementation.

All the best,
Dillon KO

PR #328, opened 2020-05-03
Avoid assert fail due to change with GitHub page

### LINKED ISSUE #327: GitHub sample fails assert
https://github.com/MechanicalSoup/MechanicalSoup/blob/424efe32972c1f3a3799fbf13456b85d073f71b8/examples/example.py#L37

and

https://github.com/MechanicalSoup/MechanicalSoup/blob/424efe32972c1f3a3799fbf13456b85d073f71b8/examples/example_manual.py#L37

This always fails for me. I've ended up changing the (first) assert to:

    assert not page.title.text.startswith('Sign in to GitHub')  # NOTE will fail for app passwords

clach04:
Fix #327

### DIFF
--- a/examples/example.py
+++ b/examples/example.py
@@ -34,11 +34,11 @@ page = browser.page
 messages = page.find("div", class_="flash-messages")
 if messages:
     print(messages.text)
-assert page.select(".logout-form")
+# NOTE assert will fail if app password is used
+assert not page.title.text.startswith('Sign in to GitHub')
 
 print(page.title.text)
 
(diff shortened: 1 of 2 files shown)


moy:
Thanks for the PR. Can you squash all commits together, and detail the issue & solution in the commit message?

hemberger:
I looked into this a little bit, and apparently the original code fails for people who have 2FA enabled for GitHub. I'm not sure what the best way to proceed is, but it's quite possible that logging into GitHub is no longer a suitable example.

PR #331, opened 2020-08-27
Fix a tiny type to `set` documenation

technicaltitch:
Fix a tiny typo in the documentation explaining how to set a file type input. (The suggested syntax sets a function call to a value.)


hemberger:
Thanks for the fix!

moy:
Indeed, thanks!

technicaltitch:
Pleasure, thanks very much for the handy package, use it for all sorts of things!

PR #333, opened 2020-09-30
Improve Docs

cheekudeveloper:


moy:
Hi,

Thank you for your interest in MechanicalSoup, but I disagree with the proposed changes. I'm not a native speaker, but I'm pretty sure all the changes you suggest would actually add errors rather than fix any of them.

Thanks,

PR #339, opened 2020-10-14
Add get_request_kwargs to check before requesting

kumarstack55:
When we use mechanicalsoup, we sometimes want to verify a request before submitting it.

If you merge this pull request, the package will be able to provide a way for the package's users to review the request.

This is my first pull request for this project. Please let me know if I'm missing anything.

### DIFF
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -139,8 +139,18 @@ class Browser(object):
         Browser.add_soup(response, self.soup_config)
         return response
 
-    def _request(self, form, url=None, **kwargs):
-        """Extract input data from the form to pass to a Requests session."""
+    @staticmethod
+    def _get_request_kwargs(method, url, **kwargs):
+        """This method exists to raise a TypeError when a method or url is
+        specified in the kwargs.
+        """
+        request_kwargs = {"method": method, "url": url}
+        request_kwargs.update(kwargs)
+        return request_kwargs
+
+    @classmethod
+    def get_request_kwargs(cls, form, url=None, **kwargs):
+        """Extract input data from the form."""
         method = str(form.get("method", "get"))
         action = form.get("action")
         url = urllib.parse.urljoin(url, action)
@@ -240,7 +250,12 @@ class Browser(object):
 
             files = DictThatReturnsTrue()
 
-        return self.session.request(method, url, files=files, **kwargs)
+        return cls._get_request_kwargs(method, url, files=files, **kwargs)
+
+    def _request(self, form, url=None, **kwargs):
+        """Extract input data from the form to pass to a Requests session."""
+        request_kwargs = Browser.get_request_kwargs(form, url, **kwargs)
+        return self.session.request(**request_kwargs)
 
     def submit(self, form, url=None, **kwargs):
         """Prepares and sends a form request.
--- a/tests/test_browser.py
+++ b/tests/test_browser.py
@@ -38,6 +38,40 @@ def test_submit_online(httpbin):
     assert 'MechanicalSoup' in json["headers"]["User-Agent"]
 
 
+def test_get_request_kwargs(httpbin):
+    """Return kwargs without a submit"""
+    browser = mechanicalsoup.Browser()
+    page = browser.get(httpbin + "/forms/post")
+    form = page.soup.form
+    form.find("input", {"name": "custname"})["value"] = "Philip J. Fry"
+    request_kwargs = browser.get_request_kwargs(form, page.url)
+    assert "method" in request_kwargs
+    assert "url" in request_kwargs
+    assert "data" in request_kwargs
+    assert ("custname", "Philip J. Fry") in request_kwargs["data"]
+
+
+def test_get_request_kwargs_when_method_is_in_kwargs(httpbin):
+    """Raise TypeError exception"""
+    browser = mechanicalsoup.Browser()
+    page = browser.get(httpbin + "/forms/post")
+    form = page.soup.form
+    kwargs = {"method": "post"}
+    with pytest.raises(TypeError):
+        browser.get_request_kwargs(form, page.url, **kwargs)
+
+
+def test_get_request_kwargs_when_url_is_in_kwargs(httpbin):
+    """Raise TypeError exception"""
+    browser = mechanicalsoup.Browser()
+    page = browser.get(httpbin + "/forms/post")
+    form = page.soup.form
+    kwargs = {"url": httpbin + "/forms/post"}
+    with pytest.raises(TypeError):
+        # pylint: disable=redundant-keyword-arg
+        browser.get_request_kwargs(form, page.url, **kwargs)
+
+
 def test__request(httpbin):
     form_html = """
     <form method="post" action="{}/post">


mechanicalsoup/browser.py
```diff
@@ -240,7 +240,14 @@ def __bool__(self):
 
             files = DictThatReturnsTrue()
 
-        return self.session.request(method, url, files=files, **kwargs)
+        request_kwargs = {"method": method, "url": url, "files": files}
+        request_kwargs.update(kwargs)
```
hemberger, line 244: If the user provides `method` or `url` arguments to `**kwargs` in our API, it would currently raise a `TypeError` due to a repeated argument passed to `session.request` (I'm working on a unit test to demonstrate this now). However, in this patch, those keys would be overwritten by the `request_kwargs` dict update. It seems like it would be a little clunky to manually check for those keys, but maybe that's what's needed here to retain the expected behavior?

hemberger (changes requested):
Hi, and thanks for your contribution to MechanicalSoup!

In principle, I think this is a fine change. I've recommended one small spelling error and identified a corner case that probably needs to be addressed. Once those are fixed, I will be ready to approve this. Then if the other maintainer approves as well we can merge it in. :)

mechanicalsoup/browser.py
```diff
@@ -139,8 +139,8 @@ def post(self, *args, **kwargs):
         Browser.add_soup(response, self.soup_config)
         return response
 
-    def _request(self, form, url=None, **kwargs):
-        """Extract input data from the form to pass to a Requests session."""
+    @classmethod
+    def get_reuqest_kwargs(cls, form, url=None, **kwargs):
```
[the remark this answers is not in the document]
kumarstack55: Thank you for pointing this out. It has been corrected.

mechanicalsoup/browser.py
kumarstack55: I hadn't thought about TypeError. Thank you for pointing that out.

It looks like we can check the key for backwards compatibility. There seems to be more than one way to implement it.

Which of the following implementation methods do you prefer?

(a) Raise TypeError in the package
(b) Or create a class method, e.g. `_get_request_kwargs()` that only merges the kwargs so that Python raises a TypeError.

kumarstack55: Since we have difficulty thinking about the msg of TypeError(msg), I thought it might be more desirable for Python to raise a TypeError.

kumarstack55:
The group of jobs in travisci are listed as Queued, but when I click on Details to check, it looks like all the jobs are successful.

mechanicalsoup/browser.py
kumarstack55: 536ce4a398c88a3d4bdf26da19502619ebb2de67 Resolve this issue.

kumarstack55:
> The group of jobs in travisci are listed as Queued, but when I click on Details to check, it looks like all the jobs are successful.

@hemberger Is there anything else I should do? :)

moy:
Commits should be squashed and the typo should be fixed in the commits summary (I just fixed it in the PR title). We keep a clean history hence a PR should include several commits only if the change needs to be split into several commits, not as leftover of previous mistakes.

mechanicalsoup/browser.py
```diff
@@ -139,8 +139,16 @@ def post(self, *args, **kwargs):
         Browser.add_soup(response, self.soup_config)
         return response
 
-    def _request(self, form, url=None, **kwargs):
-        """Extract input data from the form to pass to a Requests session."""
+    @classmethod
+    def _get_request_kwargs(cls, method, url, **kwargs):
```
moy, line 143: `cls` is unused here, right? Why not use `@staticmethod` instead of `@classmethod` then?

mechanicalsoup/browser.py
```diff
@@ -139,8 +139,16 @@ def post(self, *args, **kwargs):
...
+        # This method exists to raise a TypeError when a method or url is
+        # specified in the kwargs.
```
moy, line 145: Shouldn't this comment be a docstring?

tests/test_browser.py
```diff
@@ -38,6 +38,48 @@ def test_submit_online(httpbin):
...
+    """Raise TypeError exception"""
+    browser = mechanicalsoup.Browser()
+    page = browser.get(httpbin + "/forms/post")
+    form = page.soup.form
+    kwargs = {"method": "post"}
+    type_error_raised = None
+    try:
+        browser.get_request_kwargs(form, page.url, **kwargs)
+    except TypeError:
+        type_error_raised = True
```
moy, line 64: Please use [pytest.raises](https://docs.pytest.org/en/stable/reference.html?highlight=raises#pytest.raises) to assert that an exception is raised.

tests/test_browser.py
```diff
@@ -38,6 +38,48 @@ def test_submit_online(httpbin):
...
+    browser = mechanicalsoup.Browser()
+    page = browser.get(httpbin + "/forms/post")
+    form = page.soup.form
+    kwargs = {"url": httpbin + "/forms/post"}
+    type_error_raised = None
+    try:
+        # pylint: disable=redundant-keyword-arg
+        browser.get_request_kwargs(form, page.url, **kwargs)
+    except TypeError:
+        type_error_raised = True
```
moy, line 79: Likewise.

moy (changes requested):
On overall, this looks good, but a few more nits to be fixed before this can be merged.

mechanicalsoup/browser.py
kumarstack55: Actually, this is my first time writing a docstring.

I'll be referring to other methods to write docstrings. Please let me know if you have any concerns.

tests/test_browser.py
kumarstack55: Thanks, I'll rewrite it using pytest.raises.

tests/test_browser.py
kumarstack55: Thanks, I'll rewrite it using pytest.raises.

mechanicalsoup/browser.py
kumarstack55: Thanks for the good point. I'll use `@staticmethod` to avoid making `cls` useless.

kumarstack55:
Originally, this pull request was intended to implement `_get_request_kwargs()`, but the name was changed and `get_request_kwargs()` is the current name.

Therefore, please let me change the title of the Pull Request.

kumarstack55:
> Commits should be squashed and the typo should be fixed in the commits summary (I just fixed it in the PR title).

I squashed the commits to one. :sparkles:

moy:
On overall this looks good to me. The commit message would deserve a bit more love I think (explain why the change is needed and good).

moy (approved):

kumarstack55:
> On overall this looks good to me. The commit message would deserve a bit more love I think (explain why the change is needed and good).

Let me make the commit log more understandable and lovely. I don't plan to make any changes to the code at this time.

kumarstack55:
@hemberger @moy Let me know if there's anything else I need to do to get the pull request merged.

mechanicalsoup/browser.py
```diff
@@ -139,8 +139,17 @@ def post(self, *args, **kwargs):
...
+    def _get_request_kwargs(method, url, **kwargs):
+        """This method exists to raise a TypeError when a method or url is
+        specified in the kwargs.
+        """
+        request_kwargs = {"method": method, "url": url}
+        request_kwargs.update(kwargs)
+        return request_kwargs
+
+    @classmethod
+    def get_request_kwargs(cls, form, url=None, **kwargs):
```
hemberger, line 152: Since this is a public method, it will need a docstring.

hemberger (changes requested):
Overall, the changes look great. I've identified one simple change I'd like to see, but then I think this is ready to be merged. Thanks!

mechanicalsoup/browser.py
moy: Hmm, and this should be a @staticmethod, not @classmethod, right?

hemberger: I think they used classmethod because it calls
```python
return cls._get_request_kwargs(method, url, files=files, **kwargs)
```
However, this could be `Browser._get_request_kwargs`. I'm happy with it changed or as-is.

moy: OK, I'm fine with both too.

kumarstack55: I added a simple docstring to the public method.

hemberger:
Thanks again for your contribution!

kumarstack55:
Thank you for merging. :smile:

By the way, it occurs to me that this was the first time in my life that I've ever had my code reviewed.
I don't have a job writing programs. Furthermore, I had never even written a docstring before.
(I do some scraping as a hobby, so having Mechanicalsoup available would be very helpful. )

For me, all of the points you @hemberger @moy have given me are excellent. :sparkles:
I'm sure it would be very difficult for you to review my code.
Thanks to your thoughtful comments, I managed to get it merged. 

Again, thank you for the series of responses and for merging my request. :+1:

moy:
Thanks for the feedback. In another life, I'm a teacher, and that's what I try to explain my students: code review is not "I went through it and it doesn't look too bad", but can be much more picky *and constructive* than that ;-).

Issue #340, opened 2020-10-14
Make a release


ewjoachim:
The default documentation on ReadTheDocs points to the latest build and not the stable one. This means that following the documentation will lead to non-working examples. I can switch back to the latest stable, but I'm quite fond of the new `browser.page` properties etc.

Have you planned a release at some point ? Also, if the lib has little activity, it might be the time for the 1.0.0 :)

Also, related to the fact the library seems to have little activity these days, is it just feature-complete enough that there's no need for more contributions, or is it slightly abandoned ? Would you advise projects to start using this lib ?

hemberger:
Hi, thanks for your interest in MechanicalSoup!

I'm a bit confused about your ReadTheDocs experience. When I go to https://mechanicalsoup.readthedocs.io/, it immediately redirects to https://mechanicalsoup.readthedocs.io/en/stable/. Is that not true for everyone? I thought we had configured it to default to the Stable docs.

I would consider MechanicalSoup to be feature complete. There are some edge cases that get added as needed (as you can see from the PR's), but this is a simple package that does the job it's designed to do well. Though the active development is complete, for the most part, I don't have any reason to believe that the maintainers (of which I am one) intend to abandon this project. I expect that we'll continue to address bug reports and update python compatibility.

We have planned to release 1.0.0 for a while now, and perhaps it is soon time to actually do it. Originally, there was hesitation (at least on my part) to release 1.0.0 because I wanted to change the API in a backwards-incompatible way to address some of my grievances with the original implementation (which I believe can be done much more easily before a 1.0.0 release locks in the API). This seems unlikely to happen now (see #189), so I'd say let's move forward and release. :)

ewjoachim:
Marvelous :)

Regarding the readthedocs link: I think I accessed it through either [this](https://github.com/MechanicalSoup/MechanicalSoup/blob/b6307dfea36a57a3a8a97674e72829d3f07b890e/README.rst#L121) or [this](https://github.com/MechanicalSoup/MechanicalSoup/blob/b6307dfea36a57a3a8a97674e72829d3f07b890e/README.rst#L61) link, and it's probably bad luck from me. I might do a PR to switch those to stable :)

hemberger:
@moy Would you like to make a release? I think I've addressed all outstanding issues/PRs that I needed to take action on, so I'm happy with whatever you decide to include/exclude from the 1.0 release.

moy:
[...]

PR #341, opened 2020-10-14
Link to the stable documentation, rather than the latest one

ewjoachim:
Relates to #340 (but doesn't fix it ;) )

I've left a link to the latest documentation in the badge, but the badge is in the contribution section of the README so it would make sense... Though people might still get on the wrong version of the doc. Well, we've reduced the risk :)


hemberger (approved):
This seems sensible to me, and is along the same lines as 94c0bb01a1c720bebf691156cb3bdb441d291006. Thanks for the contribution!

PR #348, opened 2021-01-05
Drop support for EOL Python 3.5

hemberger:
This also allows us to replace `.format()` with f-strings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ language: python
 cache: pip
 python:
   - pypy3
-  - 3.5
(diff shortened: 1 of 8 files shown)


hemberger:
Further motivation for this change from `pip`:
> DEPRECATION: Python 3.5 reached the end of its life on September 13th, 2020. Please upgrade your Python as Python 3.5 is no longer maintained. pip 21.0 will drop support for Python 3.5 in January 2021. pip 21.0 will remove support for this functionality.

PR #353, opened 2021-01-09
Create workflow to publish new releases on PyPI

hemberger:


moy:
I love those "turn instructions into code" commits :-).

Issue #354, opened 2021-01-27
Incorrect encoding detected when the page does not set its encoding through a meta tag


LVerneyPEReN:
Hi,

I noticed issues of incorrect encoding detected by BeautifulSoup for a webpage which sets its encoding in the HTTP headers (requests get the right encoding) but does not specify it at the HTML level through a `<meta>` tag. In my case, the page was UTF-8 encoded but BeautifulSoup4 detects it as "Windows-1252".

I noticed encoding from obtained from requests is not passed along to BeautifulSoup in MechanicalSoup: https://github.com/MechanicalSoup/MechanicalSoup/blob/c8f39e673df70c43fcea623d8c2a649fdca95f91/mechanicalsoup/browser.py#L65-L69.

I think the issue could be fixed by using something along:

```python
response.soup = bs4.BeautifulSoup(response.content, encoding=response.encoding, **soup_config)
```

but there might be additional side effects which I do not foresee.

Is there a reason for not doing so?

Thanks!
Best,

moy:
Hi,

I don't think there's a reason for not doing what you suggest. We haven't worked much on encoding in MechanicalSoup, I guess that's the explanation.

If we go for this, obviously this will also need a test.

Do you want to work on a patch?

Regards,

hemberger:
Here's an interesting suggestion from https://stackoverflow.com/a/45643551, i.e.
```python
from bs4 import BeautifulSoup
from bs4.dammit import EncodingDetector
resp = requests.get(url)
http_encoding = resp.encoding if 'charset' in resp.headers.get('content-type', '').lower() else None
html_encoding = EncodingDetector.find_declared_encoding(resp.content, is_html=True)
encoding = html_encoding or http_encoding
soup = BeautifulSoup(resp.content, 'lxml', from_encoding=encoding)
```
I think something like this might be needed to account for the case where the encoding is in the HTML, but not in the headers.

LVerneyPEReN:
@hemberger Indeed! I'm wondering if there could be cases where HTTP request encoding (as returned by HTTP headers / `requests.encoding`) differ from the HTML encoding (as defined in the `<meta>` tag) though?

hemberger:
I imagine anything is possible out in the wild. I'm not sure that it will be possible to correctly handle every kind of misconfiguration, though. Therefore, would we want to provide an interface for enforcing a specific encoding?

PR #355, opened 2021-01-29
Use encoding from the HTTP request when no HTML encoding is specified

LVerneyPEReN:
Hi,

Here is a PR for fixing #354, along with tests. Httpbin offers a testing endpoint defining encoding at the HTTP level, but not through a HTML tag (precisely the error initially described in #354). I did not find any testing endpoints to use for other cases ("no HTTP encoding, HTML one" and "both of them").

Best,

### DIFF
--- a/tests/test_browser.py
+++ b/tests/test_browser.py
@@ -7,6 +7,11 @@ import os
 from requests.cookies import RequestsCookieJar
 import pytest
 
+from utils import (
+    prepare_mock_browser,
+    mock_get
+)
+
 
 def test_submit_online(httpbin):
     """Complete and submit the pizza form at http://httpbin.org/forms/post """
@@ -292,6 +297,55 @@ def test_post(httpbin):
     data = {'color': 'blue', 'colorblind': 'True'}
     resp = browser.post(httpbin + "/post", data)
     assert resp.status_code == 200 and resp.json()['form'] == data
+
+
+@pytest.mark.parametrize("http_html_expected_encoding", [
+    pytest.param((None, 'utf-8', 'utf-8')),
+    pytest.param(('utf-8', 'utf-8', 'utf-8')),
+    pytest.param(('utf-8', None, 'utf-8')),
+    pytest.param(('utf-8', 'ISO-8859-1', 'utf-8')),
+])
+def test_encoding(httpbin, http_html_expected_encoding):
+    http_encoding = http_html_expected_encoding[0]
+    html_encoding = http_html_expected_encoding[1]
+    expected_encoding = http_html_expected_encoding[2]
+
+    url = 'mock://encoding'
+    text = (
+        '<!doctype html>'
+        + '<html lang="fr">'
+        + (
+            (
+                '<head><meta charset="'
+                + html_encoding
+                + '"><title>Titleéàè</title></head>'
+            ) if html_encoding
+            else ''
+        )
+        + '<body></body>'
+        + '</html>'
+    )
+
+    browser, adapter = prepare_mock_browser()
+    mock_get(
+        adapter,
+        url=url,
+        reply=(
+            text.encode(http_encoding)
+            if http_encoding
+            else text.encode("utf-8")
+        ),
+        content_type=(
+            'text/html'
+            + (
+                ';charset=' + http_encoding
+                if http_encoding
+                else ''
+            )
+        )
+    )
+    browser.open(url)
+    assert browser.page.original_encoding == expected_encoding
 
 
 if __name__ == '__main__':
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -1,5 +1,6 @@
 import requests
 import bs4
+import bs4.dammit
 import urllib
 from .form import Form
 import webbrowser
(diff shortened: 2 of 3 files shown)


LVerneyPEReN:
I should have fixed the build issues.

As a side note, I should add that without this PR, the faulty page I had was taking 6s to load, versus 0.04s and the correct encoding afterwards.

tests/test_browser.py
```diff
@@ -294,5 +294,13 @@ def test_post(httpbin):
...
 
 
+def test_encoding(httpbin):
+    browser = mechanicalsoup.Browser()
+    # Get HTML encoded in UTF-8 from httpbin
+    # (UTF-8 specified in HTTP headers, no meta tags)
+    page = browser.get(httpbin + "/html")
+    assert page.soup.original_encoding == 'utf-8'
+
+
```
moy, line 304: It's good to have a test, but that would be even better to play with `requests_mock` (see `setup_mock_browser` in `tests/utils.py`) to actually cover all cases in the `html_encoding or http_encoding` expression (both set, both unset, only one set).

moy (commented):
Thanks for your submission.

`StatefulBrowser.open_fake_page()` is mostly here for testing, but would deserve some care too (we won't have any `http_encoding`, but does the constructor of `BeautifulSoup` do the right thing?)

If possible, please squash your commits together, we don't keep distinct invalid+fixup commits in our history (but we can do that while merging if you prefer).

tests/test_browser.py
LVerneyPEReN: Thanks for the review. This should be fixed!

LVerneyPEReN:
Hi,

Thanks for the review @moy ! I should have squashed everything and fixed according to your review. Let me know if anything else is needed on my end.

Best,

tests/test_browser.py
```diff
@@ -294,5 +299,49 @@ def test_post(httpbin):
...
+        '<head><title>Titleéàè</title></head>'
+        '<body></body>'
+        '</html>'
+    )
+    mock_get(
+        adapter,
+        url=url,
+        reply=text,
+        content_type='text/html; charset=utf-8'
+    )
```
moy, line 341: I think this lacks the "both HTML and HTTP specify an encoding, but they are in conflict", which IMHO is an important one. Can you add it? (that's an argument in favor of parameterization, if you refactor before, adding a case is just a matter of adding one item to the parameters list)

moy (changes requested):

tests/test_browser.py
LVerneyPEReN: I can definitely add this test, but I'm not sure what should be the correct behavior here. I would consider such conflict to be a bug on the remote side.

Not sure which one should have precedence. It is currently `html_encoding` but maybe it should be `http_encoding`?

https://github.com/MechanicalSoup/MechanicalSoup/pull/355/files#diff-8ddd0e386c398170601ea5f855d66494d04be5752fc7c88b12af0fd3bb9f48e8R75

moy: In any case, testing this is a way to document the current behavior and make sure we don't change it without noticing.

What do web browsers do in this case? I think they favor the HTTP header too (and IIRC I've already be bitten by this with my sysadmin setting the HTTP header I had no control on, and me trying to set the encoding in an HTML static page), but I didn't check.

LVerneyPEReN:
Tests are now parameterized and covering all possible cases + code is adapted to use the logic discussed in previous comment.

LVerneyPEReN:
Hi @moy, is this MR fine with the latest changes?

moy (commented):
Oops, I started commenting but forgot to submit my review :-(

I think a bit more polishing is possible, but I'm also OK with merging as-is if you want to stop working on it (and if @hemberger is OK).

tests/test_browser.py
```diff
@@ -294,5 +299,45 @@ def test_post(httpbin):
...
+    pytest.param(('utf-8', None, 'utf-8')),
+    pytest.param(('utf-8', 'ISO-8859-1', 'utf-8')),
+])
+def test_encoding(httpbin, http_html_expected_encoding):
+    url = 'mock://encoding'
+    text = (
+        '<!doctype html>'
+        + '<html lang="fr">'
+        + (
+            '<head><meta charset="utf-8"><title>Titleéàè</title></head>'
```
moy, line 314: Shouldn't this be `charset="{http_html_expected_encoding[1]}"`?

tests/test_browser.py
```diff
@@ -294,5 +299,47 @@ def test_post(httpbin):
     assert resp.status_code == 200 and resp.json()['form'] == data
 
 
+@pytest.mark.parametrize("http_html_expected_encoding", [
+    pytest.param((None, 'utf-8', 'utf-8')),
+    pytest.param(('utf-8', 'utf-8', 'utf-8')),
+    pytest.param(('utf-8', None, 'utf-8')),
+    pytest.param(('utf-8', 'ISO-8859-1', 'utf-8')),
+])
+def test_encoding(httpbin, http_html_expected_encoding):
```
moy, line 308: I'd start with
```
http_enc, html_enc, expected_enc = http_html_expected_encoding
```
and use `http_enc`, `html_enc`, ... for readability.

tests/test_browser.py
LVerneyPEReN: I'm on it, I pushed too fast :)

LVerneyPEReN:
> I think a bit more polishing is possible, but I'm also OK with merging as-is if you want to stop working on it (and if @hemberger is OK).

This is something blocking on our side, since we are dealing with webservers which are not properly defining HTTP/HTML encoding. Hence we are being confronted to encoding issues with MechanicalSoup.

Rather than maintaining our own version of the codebase, I'd rather try to get this merged ASAP and eventually a new MechanicalSoup version released to benefit everyone, if this is ok for you.

moy:
OK, the new version looks perfect. Let's merge this!

Do you need a release ASAP too?

LVerneyPEReN:
Thanks for the merge and your work on MechanicalSoup! A release ASAP would be great, if this is doable for you.

hemberger:
I'm sorry we never got around to doing a release. I'll try to get one out this week.

PR #359, opened 2021-03-14
Add a wrapper to PUT

titixbrest:
I have to use PUT method in a personnal project, but it was not exposed.

Tests passed in my env.

### DIFF
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -160,6 +160,18 @@ class Browser:
         Browser.add_soup(response, self.soup_config)
         return response
 
+    def put(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


hemberger:
Thanks for your contribution @titixbrest! Looks like the test is failing only because you're missing an extra blank line in test_browser.py:

> /home/travis/build/MechanicalSoup/MechanicalSoup/tests/test_browser.py:301:1: E302 expected 2 blank lines, found 1

Would you mind fixing that? Otherwise this LGTM!

titixbrest:
should be done ;)

hemberger (approved):
LGTM!

No need to force push. I can squash the fix when I merge. Thanks again!

PR #365, opened 2021-04-05
bugfix : Line 76, in add_soup : if 'charset' in response.headers.get(…

titixbrest:
Line 76, in add_soup : if 'charset' in response.headers.get(Content-Type) leads to TypeError: argument of type 'NoneType' is not iterable if response.headers.get(Content-Type) is not defined and Browser.__looks_like_html(response) is True

### DIFF
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -73,7 +73,7 @@ class Browser:
             # resort to bs4 sniffing, hence the special handling here.
             http_encoding = (
                 response.encoding
-                if 'charset' in response.headers.get("Content-Type")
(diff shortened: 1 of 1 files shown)


moy:
The change looks good, but the commit message should be re-worked. A line number is not relevant here (it's already in the patch, no need to repeat). I suggest

> "if 'charset' in response.headers.get(Content-Type)" expects the content-type to be a string, but fails when the content-type is not set (argument of type 'NoneType' is not iterable). Make sure we always return a string before searching for a substring.

titixbrest:
My bad ;)

moy (approved):

hemberger (approved):
Thanks for the fix!

Issue #362, opened 2021-03-28
browser.follow_link() has no way to pass kwargs to requests


johnhawkinson:
As noted elsewhere, I've recently been debugging behind an SSL proxy, which requires telling requests to not verify SSL certificates. Generally I've done that with

```python
    kwargs = { "verify": False }
    # ...
    r = br.submit_selected(**kwargs)
```

which is fine. But it's not so fine when I need to follow a link, because `browser.follow_link()` uses its `**kwargs` for BS4's tag finding, but not for actually following the link.

So instead of

```python
    r = br.follow_link(text='Link anchor', **kwargs)
```

I end up with

```python
    link = br.find_link(text='Link anchor')
    r = br.open_relative(link['href'], **kwargs)
```

I am not sure how to fix this. Some thoughts:

1. If nothing changes, add some more clarity to `browser.follow_link()`'s documentation explaining how to work around this situation.
2. Add kwargs-ish params to `browser.follow_link()`, one for BS4 and one for Requests. Of course, only one gets to be `**kwargs`, but at least one might be able to call `browser.follow_link(text='Link anchor', requests_args=kwargs)` or something.
3. Send the same `**kwargs` parameter to both

Maybe there's a better way. I guess in my case I could set this state in requests' Session object, ~which I think would be `browser.session.merge_environment_settings(...)`~ no, that's not right, I'm not sure how to accomplish it actually.

hemberger:
Hi, thanks for your interest in MechanicalSoup!

I think what you ended up with looks right to me (as that's basically all `follow_link` does), and that's probably the best you'll be able to do with the current version of MechanicalSoup.

I agree that it would be convenient to be able to forward arguments to both bs4 and requests from a single interface, and I believe we've had this issue before where we didn't know how to clearly define the forwarding endpoints. If I could remake the interface, I'd probably avoid the flat argument lists that we use now.

As far as remediation goes, your option (2) looks like our best bet, and I'd suggest that `follow_link` and `download_link` be changed in the same way for consistency. We would probably just need to handle the case of `headers` being specified like we do in `submit_selected`.

Would you be interested in submitting a PR for this change?

johnhawkinson:
> Would you be interested in submitting a PR for this change?

Yes, I'd be happy to do so later this week.

johnhawkinson:
I'm looking at this now, @hemberger , and I don't understand the reasoning for the code you referenced above:

[...]

PR #367, opened 2021-04-18
stateful_browser: self.url -> self.__state.url for internal consumers

johnhawkinson:
Be consistent and use self.__state.url over the self.url() @property for internal consumers.

Discussed somewhat in #362, the logic being that:

  self.url is an interface for external callers to get access to the
  internal state, and there's no reason to force the internal users to
  do so.

  There's extra cognitive load for readers of the code to follow
  through the indirection.

### DIFF
--- a/mechanicalsoup/stateful_browser.py
+++ b/mechanicalsoup/stateful_browser.py
@@ -12,6 +12,7 @@ from requests.structures import CaseInsensitiveDict
 
 
 class _BrowserState:
+    """An internal class whose purpose is underspecified."""
     def __init__(self, page=None, url=None, form=None, request=None):
         self.page = page
         self.url = url
@@ -68,7 +69,7 @@ class StatefulBrowser(Browser):
         # Almost same as self.form, but don't raise an error if no
         # form was selected for backward compatibility.
         self.get_current_form = lambda: self.__state.form
-        self.get_url = lambda: self.url
+        self.get_url = lambda: self.__state.url
 
     def set_debug(self, debug):
         """Set the debug mode (off by default).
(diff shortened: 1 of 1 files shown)


hemberger (commented):
Thanks for the PR! I really appreciate your dedication to helping improve this library above and beyond what you need to get it to work for you.

Looking through the code more, I see that we have a similar inconsistency with `self.page` and `self.form`. One thing to note, however, is that the `self.form` property does more than simply return `self.__state.form`: it also checks to make sure a form has been selected. Therefore, we will likely need to retain most references to `self.form`, which prevents us from using the `__state` attributes consistently, even if we were to convert `self.page`.

Does that change your evaluation of this situation at all? Maybe it hints at a deeper design flaw.

@moy Do you have any opinions on this matter?

johnhawkinson:
> Looking through the code more, I see that we have a similar inconsistency with `self.page` and `self.form`. One thing to note, however, is that the `self.form` property does more than simply return `self.__state.form`: it also checks to make sure a form has been selected. Therefore, we will likely need to retain most references to `self.form`, which prevents us from using the `__state` attributes consistently, even if we were to convert `self.page`.
> 

That's a good question!  Are those checks appropriate in these cases?
For instance, `select_form()` sets `self.__state.form` before evaluating `self.form`, so it's redundant.
It looks like all the other ones are necessary (or, at least, meaningful and probably wise): `__setitem__()`, `new_control()`, and `submit_selected()`.

I guess this means I'm wrong and we should go with `self.url` in all cases?

Or maybe we should leave well-enough-alone, I dunno!

hemberger:
I agree, there's definitely not an obvious answer here. As you noted, in some places it is important to use `self.form`, but in others it's redundant (or even incorrect, e.g. when we're setting `self.__state.form` manually in `select_form`). So it's clear we're going to have a mix no matter what we do. Perhaps the most consistent way to do it would be to distinguish between "using" and "setting", which I think would mean using the properties everywhere except in `select_form` (and obviously the `@property` methods as well).

You are certainly not responsible for making this all consistent, so if this is just a headache you want to forget, that's totally okay. :)

PR #368, opened 2021-04-18
stateful_browser: {follow,download}_link: Allow requests_kwargs

johnhawkinson:
Addresses #362: browser.follow_link() has no way to pass kwargs to requests
Accept keywords args in three ways:
  bs4_kwargs	  Explicitly passed to Beautiful Soup, via find_link, &c.
  requests_kwargs Passed to requests, via open_relative
  **kwargs	  Excess args, merged with bs4_kwargs for backwards
  		  compatibility

Adjust docstrings as appropriate.

Rename *args to *bs4_args to more clearly indicate that these positional
args go to the bs4 functions not to the requests functions.

Add tests:
test_download_link_nofile_bs4
  passes args to BeautifulSoup via bs4_kwargs
test_download_link_nofile_excess
  passes args to BeautifulSoup via excess **kwargs
test_follow_link_ua()
test_download_link_nofile_ua()
  both of which pass in requests_kwargs that set the User-Agent header

### DIFF
--- a/docs/ChangeLog.rst
+++ b/docs/ChangeLog.rst
@@ -10,6 +10,14 @@ Main changes
 
 * Dropped support for EOL Python versions: 2.7 and 3.5.
 
+* ``StatefulBrowser`` methods ``follow_link`` and ``download_link``
+  now support passing a dictionary of keyword arguments to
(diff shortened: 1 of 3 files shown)


hemberger (approved):
LGTM, thanks for your contribution!

PR #369, opened 2021-05-01
stateful_browser.py: add a response property

### LINKED ISSUE #358: Store latest response in the browser state
Hi,

It seems the latest request and soup is stored in the browser state but not the latest response object. I have a use case where this would be useful, typically for getting access to headers defined in the last response (content-type for non HTML content for example).

Would it make sense to write something to store the last response in the browser state?

Best,

hemberger:
Whenever we set the browser state, include the request response.
The `response` property will raise an `AttributeError` if the state
has not yet been set.

This is not strictly necessary, since we return the request response
in most functions that have one, but it is perhaps a convenience in
some scenarios.

Closes #358.

### DIFF
--- a/docs/ChangeLog.rst
+++ b/docs/ChangeLog.rst
@@ -10,13 +10,17 @@ Main changes
 
 * Dropped support for EOL Python versions: 2.7 and 3.5.
 
+* ``StatefulBrowser`` now has a ``response`` property that stores the
+  response related to the current browser state.
+  [`#358 <https://github.com/MechanicalSoup/MechanicalSoup/issues/358>`__]
+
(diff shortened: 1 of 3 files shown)


hemberger:
One thing that makes me vaguely hesitant about this change is users getting confused about when the response is updated. For example, `download_link` (for well-motivated reasons) does not update the internal browser state. Therefore,
```python
response = browser.download_link(...)
```
will not be the same `response` as
```python
browser.download_link(...)
response = browser.response
```
The same is already true for the current components of `_BrowserState` (page, url, and request), but at least for page and url there's a much clearer analogy with a real browser.

PR #375, opened 2021-06-09
stop submitting absolute file path for file uploads

dubbl:
I noticed that MechanicalSoup is submitting the full local file path of the uploaded file to the server, while other browsers are only sending the filename. This behavior might leak more information to the server than the user is expecting.

I checked the [HTML Spec for "File Upload state"](https://html.spec.whatwg.org/multipage/input.html#file-upload-state-(type=file)) (I think that is the applicable reference to check?), and it says:

> Filenames must not contain path components, even in the case that a user has selected an entire directory hierarchy or multiple files with the same name from different directories. Path components, for the purposes of the File Upload state, are those parts of filenames that are separated by U+005C REVERSE SOLIDUS character (\\) characters.

This PR aims to fix this by extracting the filename using `os.path.basename`, which should work on all systems. Then it only submits `filename`, and renames the old `filename` variable to `filepath`, to avoid confusions in the future.

To show the difference I set up a little Flask webserver. It just prints the uploaded file information and the user-agent, and I tested it with Firefox, Chrome, and the MechanicalSoup StatefulBrowser:
<details>
<summary>This is the log output:</summary>

```
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.77 Safari/537.36
ImmutableMultiDict([('file_input', <FileStorage: 'mech.py' ('text/plain')>)])
127.0.0.1 - - [09/Jun/2021 21:54:06] "POST /upload HTTP/1.1" 200 -
127.0.0.1 - - [09/Jun/2021 21:54:16] "GET / HTTP/1.1" 200 -
python-requests/2.25.1 (MechanicalSoup/1.1.0)
ImmutableMultiDict([('file_input', <FileStorage: '/home/dubbl/work/tmp/mech.py' (None)>)])
127.0.0.1 - - [09/Jun/2021 21:54:16] "POST /upload HTTP/1.1" 200 -
Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:89.0) Gecko/20100101 Firefox/89.0
ImmutableMultiDict([('file', <FileStorage: 'mech.py' ('text/plain')>)])
127.0.0.1 - - [09/Jun/2021 21:54:26] "POST /upload HTTP/1.1" 200 -
```
</details>

I'd be happy to get feedback on this PR, as it is my first one for this project.

### DIFF
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -1,3 +1,4 @@
+import os
 import tempfile
 import urllib
 import weakref
@@ -228,11 +229,12 @@ class Browser:
                 # If the enctype is not multipart, the filename is put in
                 # the form as a text input and the file is not sent.
                 if tag.get("type", "").lower() == "file" and multipart:
-                    filename = value
-                    if filename != "" and isinstance(filename, str):
-                        content = open(filename, "rb")
+                    filepath = value
+                    if filepath != "" and isinstance(filepath, str):
+                        content = open(filepath, "rb")
                     else:
                         content = ""
+                    filename = os.path.basename(filepath)
                     # If value is the empty string, we still pass it
                     # for consistency with browsers (see
                     # https://github.com/MechanicalSoup/MechanicalSoup/issues/250).
--- a/tests/test_browser.py
+++ b/tests/test_browser.py
@@ -155,6 +155,7 @@ def test_enctype_and_file_submit(httpbin, enctype, submit_file, file_field):
         # create a temporary file for testing file upload
         file_content = b":-)"
         pic_filedescriptor, pic_path = tempfile.mkstemp()
+        pic_filename = os.path.basename(pic_path)
         os.write(pic_filedescriptor, file_content)
         os.close(pic_filedescriptor)
         if valid_enctype:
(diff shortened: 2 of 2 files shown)


hemberger (approved):
This looks like an excellent improvement. Thanks for the fix!

moy:
Wow, indeed, there's absolutely no reason to send all this data to the server. Thanks for noticing, and thanks again for fixing.

johnhawkinson:
I spent a bit scratching my head about the whole `C:\fakepath\` business (started by "How could WHATWG seriosuly specify a standard where the Windows backslash was the delimeter of record? What century are we in…apparently this one.") but my conclusion is that we don't need to be handling anything there and that's just an API issue.

However, don't we need to handle that non-multipart case, where the filename is solely sent, if the comment is to be believed, i.e. line 243?:

https://github.com/MechanicalSoup/MechanicalSoup/blob/ff9a561938ddc5270e01cd4ee8d754f3ed4f4619/mechanicalsoup/browser.py#L229-L243

Or am I reading this wrong?

(Also, having the comment be sense-reversed from the code promotes confusion. That is, the comment discussed "If the enctype is **not** multipart" but the code's test is «If the enctype **is** multipart». Perhaps something to fix while we're here.

dubbl:
@johnhawkinson Yes, I think you are right. I didn't handle the non-multipart case.

PR #382, opened 2022-01-26
Add support for Python 3.10

hemberger:


hemberger:
There are some remaining warnings that should be fixed when the next flake8 release is published:
```
  /home/travis/virtualenv/python3.10.2/lib/python3.10/site-packages/flake8/plugins/manager.py:261: DeprecationWarning: SelectableGroups dict interface is deprecated. Use select.
    eps = importlib_metadata.entry_points().get(self.namespace, ())
```

Issue #357, opened 2021-02-16
Migrate from TravisCI to GitHub Actions?


hemberger:
As of Nov 2020, TravisCI has given open source projects on their "Free" plan 10k build credits. These are non-renewable, and we have already used 2630 of 10000 credits. Each time we trigger the tests, we use about 50 credits, so we may want to preempt the eventual message that I recently got for other projects:
> Builds have been temporarily disabled for private and public repositories due to a negative credit balance.

According to the [TravisCI FAQ](https://docs.travis-ci.com/user/billing-faq/#what-if-i-am-building-open-source), we can e-mail them to request an additional one-time OSS credit allotment, and _potentially_ get a monthly allotment once that is used up. We can try to do this, but I've heard stories about unsatisfying interactions with their support team.

Ultimately, it may be less effort to simply switch to GitHub Actions.

hemberger:
Update as of 2021-04-21:
> You have used 4430 of 10000 credits

maximino-dev:
I recently learned how to use github actions, and created a file which is very similar to the travis.yml file.
Github actions seems very easy to use and can be used in an unlimited way.

Here's what the file looks like :
`
name: Python package

on:
  push:
    branches: [ master ]
  pull_request:
    branches: [ master ]

jobs:
  build:

    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
      matrix:
        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "pypy3"]

    steps:
    - uses: actions/checkout@v2
    - name: Set up Python ${{ matrix.python-version }}
      uses: actions/setup-python@v2
      with:
        python-version: ${{ matrix.python-version }}
    - name: Install dependencies for pypy3
      if: ${{ matrix.python-version == 'pypy3' }}
      run: |
        sudo apt-get install python-dev libxml2-dev libxslt-dev libz-dev
    - name: Install dependencies
      run: |
        python -m pip install --upgrade pip
        pip install -r requirements.txt
        pip install -r tests/requirements.txt
    - name: Lint with flake8
      run: |
        flake8 --version
        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
    - name: Test with pytest
      run: |
        if [[ $TRAVIS_PYTHON_VERSION == pypy* ]]; then pytest -o 'flake8-ignore=*.py ALL'; fi
[...]

PR #384, opened 2022-02-08
Switch to Github Actions

maximino-dev:
Added the python-package.yml file to use the github actions CI instead of travis (see #357),

like in the travis.yml file, the github actions file creates environment for different python versions, install dependencies and run pytest. Codecov is triggered with the last job.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -1,0 +1,38 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Build and Test
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "pypy-3.8"]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install -r requirements.txt
+        python -m pip install -r tests/requirements.txt
+    - name: Test with pytest without flake8 (for pypy)
+      if: startsWith(matrix.python-version, 'pypy')
+      run: |
+        # We run pytest without the lint with flake8 in the pypy version, otherwise, the worflow takes too long time to execute.
+        pytest -o 'flake8-ignore=*.py ALL' 
+    - name: Test with pytest
+      if: startsWith(matrix.python-version, 'pypy') != true
+      run: |
+        pytest
+    - name: Upload Coverage to Codecov
+      uses: codecov/codecov-action@v1


.github/workflows/python-package.yml
```diff
@@ -0,0 +1,33 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Python package
```
hemberger, line 4: This could probably do with a more informative name, maybe something like "Build and Test" or "Run Tests"?

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,33 @@
...
+
+on: [push, pull_request]
```
hemberger, line 6: This could probably do with a more informative name, maybe something like "Build and Test" or "Run Tests"?

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,33 @@
...
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+        pip install -r tests/requirements.txt
+    - name: Test with pytest
+      run: |
+        if [[ $TRAVIS_PYTHON_VERSION == pypy* ]]; then pytest -o 'flake8-ignore=*.py ALL'; fi
+        if [[ $TRAVIS_PYTHON_VERSION != pypy* ]]; then pytest; fi
```
hemberger, line 31: We certainly can't refer to TravisCI-specific variables. In this case, I'd try removing the conditional entirely for now, so that we can assess if it's still needed (maybe the issue of flake8 running slowly in pypy is fixed now).

hemberger (changes requested):
This looks like a great start, thanks!

One thing that is missing from the TravisCI build is the gitter notifications. Can that be re-implemented here?

.github/workflows/python-package.yml
maximino-dev: Sorry for that, I forgot to take a look at those conditions.
Without the conditions, the workflow for the different 3.x python versions finishes in ~20s, but with pypy it lasts nearly 1m30, so I guess that the issue isn't fixed. I can add the same condition with the github actions syntax.

hemberger: Sure, that sounds good to me. Thanks for checking!

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,37 @@
...
+name: Build and Test
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
moy, line 13: Do we need this? When one of the builds fails, we'll usually focus on this one and fix it, we don't need to run the others. Faster feedback, less energy waste, and we can still activate `fail-fast` when really needed while debugging a failed build.

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,37 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+        pip install -r tests/requirements.txt
+    - name: Test with pytest without flake8 (for pypy)
+      if: ${{ matrix.python-version == 'pypy-3.8' }}
```
moy, line 29: Using `startsWith(matrix.python-version, 'pypy')` (https://docs.github.com/en/actions/learn-github-actions/expressions#startswith) will avoid forgetting to change the 3.8 part when we upgrade pypy ;-).

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
```
moy, line 26: The recommended way to run `pip` these days is `python -m pip`, which makes sure you're using the same version of Python for pip and when running python programs. See e.g. https://pip.pypa.io/en/stable/user_guide/. The `pip` command normally works, but causes headaches when `pip install foo` installs `foo` for a version of python and `python my-script.py` runs another version where `foo` isn't installed.

Let's use `python -m pip` everywhere.

moy (changes requested):
Don't forget to clean up your history in the final submission. You probably need only one commit here. At worse, we can squash your commits while merging, but let's take good habits for next time.

(You can't claim your Git teacher didn't teach you `git rebase` ;-) )

moy:
Hmm, BTW:

For a 1-commit PR, the PR description can be the same as the commit message. Include the string "Fixes #357" in your commit message so that merging the PR automatically closes the bug.

We usually use imperative mood in commit message, not past tense. See e.g. https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/

Even if it's documented in #357, remind the reader *why* this is a good change in the commit message.

.github/workflows/python-package.yml
hemberger: I wondered the same thing, but `fail-fast: false` has grown on me after some thought. Surely it depends on the failure mode, but I can imagine a fair number of scenarios where we end up wasting more energy (both human and machine) by not getting the full picture up front. For example, only seeing one failed matrix result may make it unclear if it's a systematic issue across all versions, or if it is exclusive to one particular build. Another example is if we fix one version only to rerun the workflow and find that another version is failing for a different reason. Given that our tests only take ~20 seconds to complete, I don't think there's too much waste involved with letting them all finish. I could be convinced otherwise, however. :)

moy: No strong opinion on this. One more argument in favor of not deactivating fail-fast is that it removes a line of code from our config file ;-).

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,37 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install -r requirements.txt
+        python -m pip install -r tests/requirements.txt
+    - name: Test with pytest without flake8 (for pypy)
+      if: startsWith(matrix.python-version, 'pypy')
```
moy, line 29: This deserves a comment to remind the reader why we're not running flake8 with pypy (performance, IIRC).

moy:
Your commit message documents stuff that do not exist anymore:

> Add pypy3 version in python-package.yml file
> 
> Changed name of the work flow to "Build and test" and removed conditions for flake8 in pypy
> Add condition for flake8 in pypy (like in the travis.yaml file),
> Changed details, like replacing "pip" by "python -m pip"

These are changes wrt your previous drafts, that won't end up in our history. The commit message shall document the changes compared to the parent commit.

It's important for the commit message to explain why the change is good. The reader already knows that you've added a config file for GitHub action by reading the diff, but can not necessarily guess why we want to switch.

Don't forget to add `Fixes #357` to the commit message.

moy (approved):
The only remaining point is the `fail-fast: false`, for which any option is OK for me.

.github/workflows/python-package.yml
maximino-dev: In my opinion, it's better to keep `fail-fast` as false because, it is quite nice to have an overview for the different versions, it's esaier to compare if a build fails, and like hemberger said, the test aren't very long.
But in another way, the default value is true, so indeed, we can remove the line to simplify the config file.

maximino-dev:
> This looks like a great start, thanks!
> 
> One thing that is missing from the TravisCI build is the gitter notifications. Can that be re-implemented here?

Github actions doesn't have an official webhook job yet (unlike travis), but I found some codes on github that allows to send webhook events. For example this code: _https://github.com/joelwmale/webhook-action_ which can be easily ran in the .yml file by adding : 
`- name: Gitter notification
  uses: joelwmale/webhook-action@2.1.0
  with:
    # The url to send the webhook event to
    url: https://webhooks.gitter.im/e/d591e73173ccf74fe8ba`

Another solution is to directly set up a webhook in the Github settings (Going to _Settings_ - _Wehbooks_ - _Add webhook_).

hemberger (approved):
I see a few options for workflow stuff in "Settings -> Webhooks", but nothing that gives us the fine-grained control of Travis' `on_success: change` option, same with the notify Actions I've found.

These notifications are far from essential, so I'd be happy to move forward without them for now. Perhaps in the future there will be improved Actions integration with Gitter and we can reconsider adding notifications.

hemberger:
Thanks for your work on this @maximino-dev!

PR #392, opened 2022-03-02
Prevent test workflow from triggering twice

hemberger:
Restrict the branches on which the test workflow is triggered by pushes
so that we don't trigger it twice on PR branches that are created in
the upstream repository.

In general, most developers will create the PR branch in their fork,
but this is not the case for, e.g., PRs that dependabot creates.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -3,7 +3,13 @@
 
 name: Build and Test
 
-on: [push, pull_request]
+# Run workflow for pushes to main and version branches, or pull requests
+on:
+  push:
+    branches:
+      - main
+      - 'v[0-9.]+'
+  pull_request:
 
 jobs:
   build:


hemberger:
It's a good question! The `push` and `pull_request` events behave just like Travis (that is, they do different things, as you have described). But since `push` is triggered in a PR branch only if it is pushed to the upstream repo (for both Actions and Travis), and most human devs will be submitting PRs from a fork, it seems like the main beneficiary of both triggers would be dependabot. In this case, it's mostly redundant because dependabot only submits PRs that are up-to-date with its base branch anyway.

I have no strong preferences for how we trigger our CI, but only running `pull_request` on PRs is the most common paradigm I've seen, including in some of the core Actions, e.g. [actions/checkout](https://github.com/actions/checkout/blob/main/.github/workflows/test.yml):
```yaml
on:
  pull_request:
  push:
    branches:
      - main
      - releases/*
```

PR #394, opened 2022-03-02
Speed up workflow with pip cache

hemberger:
Installing pip dependencies constitutes about half of our workflow
runtime, so using the builtin pip cache from the `setup-python` action
may make our CI faster. The workflows will need to be monitored to
assess the effectiveness of the cache.

GitHub will purge the cache if it goes unused for a certain amount of
time (presently 7 days).


hemberger:
This seems to be worth a shot. I'll keep an eye out on the build times.

PR #395, opened 2022-03-03
add GitHub URL for PyPi

andriyor:
Warehouse now uses the project_urls provided to display links in the sidebar on [this screen](https://pypi.org/project/requests/), as well as including them in API responses to help the automation tool find the source code for Requests.

Example: [pypi.org/pypi/requests/json](https://pypi.org/pypi/requests/json)
Docs: [packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,6 +49,9 @@ setup(
     long_description=read('README.rst', about['__github_url__'], about[
         '__github_assets_absoluteURL__']),
     url=about['__url__'],
+    project_urls={
+        'Source': 'https://github.com/MechanicalSoup/MechanicalSoup',
+    },
 
     license=about['__license__'],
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -49,6 +49,9 @@ def read(fname, URL, URLImage):
     long_description=read('README.rst', about['__github_url__'], about[
         '__github_assets_absoluteURL__']),
     url=about['__url__'],
+    project_urls={
+        'Source': 'https://github.com/MechanicalSoup/MechanicalSoup',
```
moy, line 53: We already have the string in `__version__.py` to make it available to the source code itself. To avoid duplicating, it should be:

```suggestion
        'Source': about['__github_url__'],
```

moy (changes requested):

hemberger (approved):
Very useful, thanks for your contribution!

PR #396, opened 2022-04-04
Add werkzeug version in tests/requirements.txt file

maximino-dev:
In the continuous integration workflow, when pytest runs and calls httpbin, there was an incompatibility issue because httpbin imports a class called 'BaseResponse' from the werkzeug library. Werkzeug recently released a new version (2.1.0) removing this class because it's deprecated. Adding the condition `werkzeug < 2.1.0` in the tests/requierements.txt file will temporarly solve the problem.
I requested a PR in the httpbin library, but this little modification will avoid issues in the workflow.

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -4,3 +4,4 @@ pytest-flake8
 pytest-httpbin
 pytest-mock
 requests_mock >= 1.3.0
+werkzeug < 2.1.0


hemberger:
Can you provide a link to the httpbin PR? Thanks!

maximino-dev:
> 
Of course, here it is : https://github.com/postmanlabs/httpbin/pull/674

hemberger (approved):
LGTM, thanks! Once a new version of httpbin is released, we should probably remove this requirement though.

moy:
Merged. It may deserve a quick release, to avoid issues for people trying to install MechanicalSoup, but I won't have time to make it shortly.

hemberger:
It only affects the dev requirements, and I would expect anyone installing it in dev mode to be using the main branch instead of a release version. However, if you still think this warrants a release, I can make one.

moy:
> It only affects the dev requirements

Ah, indeed, I forgot about that. Forget about the release, sure.

Issue #385, opened 2022-02-09
Typing annotations and typechecking with mypy or pyright?


moy:
We already have basic static analysis with flake8 (and the underlying pyflakes), but using typing annotations and a static typechecker may 1) find more bugs, 2) help our users by providing completion and other smart features in their IDE.

mypy is the historical typechecker, pyright is a more recent one which in my (very limited) experience works better (it's also the tool behind the new Python mode of VSCode). So I'd suggest pyright if we don't have arguments to choose mypy.

For now, neither tool can typecheck the project without error, so a first step would be to add the necessary annotations to get an error-free pyright check.

hemberger:
Great suggestion! I'm 100% on board with adding type annotations. I'll let this sit for a while in case anyone wants to use this as a way to get involved with MechanicalSoup, but if there are no takers for a while (a month? a year?) then I'd be happy to add them myself.

gophra:
I want to get involved with this issue. What I should do?

hemberger:
@gophra Thanks for offering to get involved! As far as I understand, type annotations can be added incrementally, so I think getting some infrastructure for _testing_ the type annotations would be important. I see a lot of unofficial GitHub Actions for both pyright and mypy, so it'd be really useful to get something like that set up.

Alternatively, if you aren't comfortable working with GitHub Actions, starting to typehint a few functions would still be a great start. How you contribute is totally up to you!

hemberger:
@Daniel-Barbera That sounds great to me! I think you'll probably need to start from scratch, bearing in mind what we concluded in https://github.com/MechanicalSoup/MechanicalSoup/pull/398#issuecomment-1126200105. That does mean there will probably need to be some changes beyond just adding typehints, so feel free to bring up any questions you have here for discussion. Thanks!

PR #398, opened 2022-05-07
Add typehints

gophra:


moy:
I had a quick look and it looks good to me (not a real review yet). Note that in this project, we try to keep a clean history hence every commit should have a nice commit message. If possible, please squash your commits before we merge, otherwise we'll do that while merging.

hemberger (approved):
LGTM, thanks for your contribution!

moy (commented):
The PR is not really wrong per-se, but adds new warnings from `mypy` and `pyright`.

Adding typehints can be done with two goals in mind:
1. Find bugs or help our IDE provide relevant feedback in our own codebase
2. Find bugs or help IDEs provide relevant feedback for users of MechanicalSoup

Having more warnings with than without the PR is counter-productive wrt objective 1., but still valid for 2.

An example of new warning:

```python
    def add_soup(response: requests.Response, soup_config):
            ...
            response.soup = bs4.BeautifulSoup(
                ...
```
Now that the typechecker knows that `response`'s type, it knows that the `.soup` attribute doesn't exist, and complains. See e.g. https://stackoverflow.com/questions/61213745/typechecking-dynamically-added-attributes for a solution.

Unfortunately, I believe users of MechanicalSoup will have the same issue if they access the `.soup` attribute.

I think we should at least discuss this in the commit message.

Then, I see two ways to proceed:
A. decide that this PR in its current state is counter-productive and wait for (or write ;-) ) a better PR.
B. consider this PR as a good first step, and merge it (with an improved commit message).

I'd vote for B. (and in any case, A. should not be considered as a discouragement !).

hemberger:
I was afraid that might be the case. We probably should have made the response + soup object a new class, rather than hijack the response object. I guess that's one of the inherent dangers with Python's duck typing framework -- it can be challenging to shoehorn traditional typing in after the fact (though I think it's a very worthy goal!).

Please correct me if I'm wrong, but I don't think mypy will actually use any of our typehints (when MechanicalSoup is an imported library) until we provide stubs. In that case, the presence of the typehints shouldn't introduce any warnings into user projects yet. So I'm fine with merging as-is, _but_ we may want to first make sure this problem can even be solved...

moy:
> Please correct me if I'm wrong, but I don't think mypy will actually use any of our typehints (when MechanicalSoup is an imported library) until we provide stubs.

That seems true with mypy. Pyright (more recent, and in my experience better) has a slightly different behavior: it will use typing annotations if mechanicalsoup is imported from a subdirectory (i.e. you copied the `mechanicalsoup` directory in your project) or if the mechanicalsoup directory contains a [`py.typed`](https://dev.to/whtsky/don-t-forget-py-typed-for-your-typed-python-package-2aa3) file (possibly empty, it's just a marker saying "do consider type here").

For example:

```
$ cat test.py                                                             
import mechanicalsoup

browser = mechanicalsoup.Browser()
page = browser.get("http://example.com")
reveal_type(page)
$ pyright test.py 
No configuration file found.
No pyproject.toml file found.
stubPath /tmp/test/typings is not a valid directory.
Assuming Python platform Linux
Searching for source files
Found 1 source file
/tmp/test/test.py
  /tmp/test/test.py:5:13 - info: Type of "page" is "Unknown"
0 errors, 0 warnings, 1 info 
Completed in 0.809sec
$ touch /home/mmoy/.local/lib/python3.8/site-packages/mechanicalsoup/py.typed
$ pyright test.py                                                            
No configuration file found.
No pyproject.toml file found.
stubPath /tmp/test/typings is not a valid directory.
Assuming Python platform Linux
Searching for source files
Found 1 source file
/tmp/test/test.py
  /tmp/test/test.py:5:13 - info: Type of "page" is "Response"
0 errors, 0 warnings, 1 info 
Completed in 0.915sec
```

In other words, until we add a `py.typed` file to our install, we can safely add type hints without bothering our users in case our hints are too strict.

FWIW, mypy apparently also needs this `py.typed` but also needs a stub file while `pyright` can extract type directly inlined into the code:

```
$ mypy test.py
test.py:5: note: Revealed type is "Any"
```

That's a good argument in favor of B I believe.

hemberger:
Coming back to this much later, I think I agree that it would be better to make sure we're not introducing errors in mypy/pyright. Looks like there's some new action on #385, so I'll close this for now. Thank you for your contribution!

PR #399, opened 2022-05-14
Support associated elements with a form

maximino-dev:
According to the [HTML spec](https://html.spec.whatwg.org/multipage/forms.html#category-listed), a list of elements (button, input, object, output, select, textarea) can have a form attribute, so they can be placed anywhere on the page and be associated with a form without being a child of it. MechanicalSoup don't support this, so I added a code that scraps the page and finds all associated element of a selected form using beautiful soup methods.

### DIFF
--- a/mechanicalsoup/stateful_browser.py
+++ b/mechanicalsoup/stateful_browser.py
@@ -207,10 +207,29 @@ class StatefulBrowser(Browser):
         :return: The selected form as a soup object. It can also be
             retrieved later with the :attr:`form` attribute.
         """
+
+        def find_associated_elements(form_id):
+            """Find all elements associated to a form
+                (i.e. an element with a form attribute -> ``form=form_id``)
+            """
+
+            # Elements which can have a form owner
+            elements_with_owner_form = ("input", "button", "fieldset",
+                                        "object", "output", "select",
+                                        "textarea")
+
+            found_elements = []
+
+            for element in elements_with_owner_form:
+                found_elements.extend(
+                    self.page.find_all(element, form=form_id)
+                )
+            return found_elements
+
         if isinstance(selector, bs4.element.Tag):
             if selector.name != "form":
                 raise LinkNotFoundError
-            self.__state.form = Form(selector)
+            form = selector
         else:
             # nr is a 0-based index for consistency with mechanize
             found_forms = self.page.select(selector,
@@ -220,7 +239,15 @@ class StatefulBrowser(Browser):
                     print('select_form failed for', selector)
                     self.launch_browser()
                 raise LinkNotFoundError()
-            self.__state.form = Form(found_forms[-1])
+
+            form = found_forms[-1]
+
+        if form and form.has_attr('id'):
+            form_id = form["id"]
+            new_elements = find_associated_elements(form_id)
+            form.extend(new_elements)
+
+        self.__state.form = Form(form)
 
         return self.form
 
--- a/tests/test_stateful_browser.py
+++ b/tests/test_stateful_browser.py
@@ -426,6 +426,30 @@ def test_select_form_tag_object():
             browser.select_form(soup.find("p"))
 
 
+def test_select_form_associated_elements():
+    """Test associated elements outside the form tag"""
+    forms = """<form id="a"><input><textarea></form><input form="a">
+               <textarea form="a"/><input form="b">
+               <form id="ab" action="/test.php"><input></form>
+               <textarea form="ab"></textarea>
+            """
+    with mechanicalsoup.StatefulBrowser() as browser:
+        browser.open_fake_page(forms)
+        elements_form_a = set([
+            "<input/>", "<textarea></textarea>",
+            '<input form="a"/>', '<textarea form="a"></textarea>'])
+        elements_form_ab = set(["<input/>", '<textarea form="ab"></textarea>'])
+        form_by_str = browser.select_form("#a")
+        form_by_tag = browser.select_form(browser.page.find("form", id='a'))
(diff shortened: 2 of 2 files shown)


hemberger:
Thanks for your contribution! It looks like this is a great start, but I have a couple things that I think could be improved:
1. This should work whether we pass a string selector or a `bs4.element.Tag` to `select_form`. Right now it only works with a string selector.
2. You're using `selector[1:]` to identify the form, but that really makes a lot of assumptions about what `selector` can be. It won't always be something as simple as `#a`.

I would recommend moving this logic outside the if-else branch of `select_form`, and then use the fact that an external input element can only have the `form` attribute if the form has an `id`. So, independent of the input selector, grab the `id` from the selected form and then, if it exists, find all form-associated elements with `form=<form_id>`.

It might be a good idea to add a few more test cases to cover the various scenarios that the current code does not work with (more complex string selectors and `bs4.element.Tag` selectors).

Thanks again!

maximino-dev:
Thank you for your fast comment, and review ! 
I think that indeed your solution is more efficient and better to handle complex string selectors, so I rearranged the code and added more tests, now it works for a `bs4.element.Tag` too.

mechanicalsoup/stateful_browser.py
```diff
@@ -207,14 +207,46 @@ def select_form(self, selector="form", nr=0):
...
         else:
             # nr is a 0-based index for consistency with mechanize
             found_forms = self.page.select(selector,
                                            limit=nr + 1)
+
+            if len(found_forms) > 0 and found_forms[-1].has_attr('id'):
+                form_id = found_forms[-1]["id"]
+                new_elements = find_associated_elements(form_id)
+                found_forms[-1].extend(new_elements)
+
```
hemberger, line 249: The changes are looking good, thanks! I have one small suggestion: what if we refactor this a bit to reduce some code duplication? I'm thinking something along the lines of:

```python
if isinstance(selector, bs4.element.Tag):
    ...
    form = selector
else:
    ...
    form = found_forms[-1]

if form.has_attr("id"):
    new_elements = find_associated_elements(form["id"])
    form.extend(new_elements)

self.__state.form = Form(form)

return self.form
```

moy (changes requested):

maximino-dev:
Thank you for all your comments, it helps me improving my skills, I try to learn from that.
I have modified the code, tell me if there's something left to change.

mechanicalsoup/stateful_browser.py
```diff
@@ -207,20 +207,52 @@ def select_form(self, selector="form", nr=0):
...
             found_forms = self.page.select(selector,
                                            limit=nr + 1)
+
+            if len(found_forms) > 0:
+                form = found_forms[-1]
+            else:
+                form = None
+
             if len(found_forms) != nr + 1:
                 if self.__debug:
```
moy, line 245: You want that error-checking code to be right after the call to `self.page.select(...)`, hence insert your code below this rather than above.

tests/test_stateful_browser.py
```diff
@@ -426,6 +426,30 @@ def test_select_form_tag_object():
...
+    with mechanicalsoup.StatefulBrowser() as browser:
+        browser.open_fake_page(forms)
+        elements_form_a = set([
+            "<input/>", "<textarea></textarea>",
+            '<input form="a"/>', '<textarea form="a"></textarea>'])
+        elements_form_ab = set(["<input/>", '<textarea form="ab"></textarea>'])
+        form_by_str = browser.select_form("#a")
+        form_by_tag = browser.select_form(browser.page.find("form", id='a'))
+        form_by_css = browser.select_form("form[action$='.php']")
+        assert set(list(map(str, form_by_str.form.find_all((
```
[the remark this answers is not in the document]
maximino-dev: I understand, maybe a list comprehension is easier to read, I use it sometimes when I need to filter data, like `[ x for x in foo if x == ... ]`, otherwise I use map, but I can change it anyway.

maximino-dev:
> FYI, we receive one notification email for each push, so when you do multiple pushes to fix PEP8 we get one notification for every try. Better run flake8 locally, it's faster for you and less noisy for project maintainers.

Sorry for that, pushing and running on the Github CI was easier for me beacuse I didn't need to create a virtual env and install all packages, but I didn't know for the notifications.

hemberger:
@maximino-dev Just checking in since it's been a few weeks. Is this code ready for re-review?

maximino-dev:
> @maximino-dev Just checking in since it's been a few weeks. Is this code ready for re-review?

Yes it is, I modified the code taking into account the requested changes.

mechanicalsoup/stateful_browser.py
```diff
@@ -207,20 +207,49 @@ def select_form(self, selector="form", nr=0):
...
+            elements_with_owner_form = ("input", "button", "fieldset",
+                                        "object", "output", "select",
+                                        "textarea")
+
+            found_elements = []
+
+            for element in elements_with_owner_form:
+                for new_element in self.page.find_all(element,
+                                                      form=form_id):
+                    found_elements.append(new_element)
```
moy, line 226: Isn't this a verbose way to say
```
fond_elements.extend(
    self.page.find_all(element, form=form_id)
)
```
?

moy (commented):
Other than the minor style nits, this looks good to me.

moy (approved):
It seems the PR is ready for merging. Anything left on the TODO-list?

hemberger (approved):
LGTM! Thanks for all your work on this. It is much appreciated.

PR #400, opened 2022-06-07
Fix test_follow_link_ua to work with newer httpbin

sfreilich:
Use the compatibility function open_legacy_httpbin in test_follow_link_ua so that it works with newer versions of httpbin.


hemberger (approved):
Fine with me, thanks!

hemberger:
@moy and I are co-maintainers, so I like to give him a few days to respond before I merge anything. I'll check back soon, and if there are no further comments I'll do the merge. Thanks again!

PR #401, opened 2022-06-07
Return response in open_legacy_httpbin

sfreilich:
That makes it a better drop-in replacement for browser.open


hemberger (approved):
Sure, seems reasonable to me. Thanks!

PR #404, opened 2022-07-02
Fix typos and remove trailing spaces

kianmeng:


moy (approved):

hemberger (approved):
Thanks for your patch! The test failures look like they're related to module updates rather than anything having to do with the changes here.

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #409, opened 2022-11-10
Fix CodeQL warnings

hemberger:
* Always use context managers for file resources. Fixes rule py/file-not-closed.

* Don't assign a function return value to a variable when asserting that the function will raise an Exception. Fixes rule py/multiple-definition.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,13 +8,15 @@ from setuptools import setup  # Always prefer setuptools over distutils
 
 def requirements_from_file(filename):
     """Parses a pip requirements file into a list."""
-    return [line.strip() for line in open(filename, 'r')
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -8,13 +8,15 @@
 
 def requirements_from_file(filename):
     """Parses a pip requirements file into a list."""
-    return [line.strip() for line in open(filename, 'r')
-            if line.strip() and not line.strip().startswith('--')]
+    with open(filename, 'r') as fid:
```
moy, line 11: Just a nitpick, but I'd call the variable `fd` for File Descriptor, not `fid`ŧ

hemberger: Sounds good, I've made this change. Thanks!

PR #411, opened 2022-12-06
Add support for no button to `StatefulBrowser.submit_selected`

### LINKED ISSUE #403: Submit form with no button
Some web pages use JavaScript postbacks to submit forms. This means that a submitted form has nothing to do with which button is pressed. In fact, including form data for the "pressed" button can mess things up in such a case.

Accordingly, it would be great to have a way of specifying that *no* button "press" should be simulated when calling `submit_selected`. Right now, `btnName = None` means that the first "submit" button is auto-selected — but there is no way to specify that *no* submit button press should be simulated. Perhaps another (boolean) parameter could be used for this? The `Form.choose_submit` method would also need to be enhanced accordingly.

alexreg:
Resolves #403

### DIFF
--- a/mechanicalsoup/form.py
+++ b/mechanicalsoup/form.py
@@ -306,9 +306,10 @@ class Form:
     def choose_submit(self, submit):
         """Selects the input (or button) element to use for form submission.
 
-        :param submit: The bs4.element.Tag (or just its *name*-attribute) that
-            identifies the submit element to use. If ``None``, will choose the
-            first valid submit element in the form, if one exists.
+        :param submit: The :class:`bs4.element.Tag` (or just its
+            *name*-attribute) that identifies the submit element to use. If
+            ``None``, will choose the first valid submit element in the form,
+            if one exists.
 
         To simulate a normal web browser, only one submit element must be
         sent. Therefore, this does not need to be called if there is only
--- a/mechanicalsoup/form.py
+++ b/mechanicalsoup/form.py
@@ -309,7 +309,8 @@ class Form:
         :param submit: The :class:`bs4.element.Tag` (or just its
             *name*-attribute) that identifies the submit element to use. If
             ``None``, will choose the first valid submit element in the form,
-            if one exists.
+            if one exists. If ``False``, will not use any submit element;
+            this is useful for simulating AJAX requests, for example.
 
         To simulate a normal web browser, only one submit element must be
         sent. Therefore, this does not need to be called if there is only
@@ -362,7 +363,7 @@ class Form:
                 # omitted from the submitted form data.
                 del inp['name']
 
-        if not found and submit is not None:
+        if not found and submit is not None and submit is not False:
             raise LinkNotFoundError(
                 f"Specified submit element not found: {submit}"
             )
--- a/mechanicalsoup/stateful_browser.py
+++ b/mechanicalsoup/stateful_browser.py
@@ -268,12 +268,17 @@ class StatefulBrowser(Browser):
 
         :return: Forwarded from :func:`Browser.submit`.
 
-        If there are multiple submit input/button elements, passes ``btnName``
-        to :func:`Form.choose_submit` on the current form to choose between
-        them. If `update_state` is False, form will be submitted but the
-        browser state will remain unchanged. This is useful for forms that
-        result in a download of a file. All other arguments are forwarded to
-        :func:`Browser.submit`.
+        :param btnName: Passed to :func:`Form.choose_submit` to choose the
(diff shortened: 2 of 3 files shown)


tests/test_stateful_browser.py
```diff
@@ -170,6 +170,39 @@ def test_submit_btnName(expected_post):
...
+            ('cancel', 'Cancel'),
+        ], id='input'),
+    pytest.param(
+        [
+            ('text', '= Heading =\n\nNew page here!\n'),
+            ('comment', 'Selecting a button submit'),
+            ('preview', 'Preview Page'),
+            ('diff', 'Review Changes'),
+            ('save', 'Submit changes'),
+            ('cancel', 'Cancel'),
```
hemberger, line 190: We may need to rethink the implementation of this change, because it posts _all_ of the submit elements, rather than none of them:

```html
        <input type="submit" name="preview" value="Preview Page" />
        <input type="submit" name="diff" value="Review Changes" />
        <input type="submit" id="save" name="save" value="Submit changes" />
        <button type="submit" name="cancel" value="Cancel" />
```
I assume this is not the intended behavior? If we don't want any of the submit elements to be selected, then we probably need to modify `form.choose_submit` to delete all submit inputs if `submit is False` (or perhaps a new method with some refactoring).

hemberger (changes requested):
Thanks for the PR! This looks like what we discussed in #403, except it seems to contain some unrelated changes. Maybe you pushed some commits you did not intend to push?

mechanicalsoup/stateful_browser.py
```diff
@@ -168,6 +168,11 @@ def open_relative(self, url, *args, **kwargs):
         """
         return self.open(self.absolute_url(url), *args, **kwargs)
 
+    def update_state(self, resp):
+        """Set the Browser's state to the given response."""
+        self.__state = _BrowserState(page=resp.soup, url=resp.url,
+                                     request=resp.request)
```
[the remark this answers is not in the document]
alexreg: This was an accident, sorry.

tests/test_stateful_browser.py
alexreg: Yeah, I was wondering about that... Wasn't sure what the behaviour ought to be, but that's fair enough. I'll have a look into it.

alexreg:
@hemberger Thanks for having a look. I've addressed your comments, and the current solution may more along the right lines.

tests/test_stateful_browser.py
```diff
@@ -170,6 +170,31 @@ def test_submit_btnName(expected_post):
...
+            ('comment', 'Selecting an input submit'),
+        ], id='input'),
+    pytest.param(
+        [
+            ('text', '= Heading =\n\nNew page here!\n'),
+            ('comment', 'Selecting a button submit'),
+        ], id='button'),
+])
+def test_submit_no_btn(expected_post):
+    '''Tests that the btnName argument chooses the submit button.'''
```
hemberger, line 186:
```suggestion
    '''Tests that no submit inputs are posted when btnName=False.'''
```

mechanicalsoup/form.py
```diff
@@ -306,9 +306,11 @@ def new_control(self, type, name, value, **kwargs):
...
         """Selects the input (or button) element to use for form submission.
 
-        :param submit: The bs4.element.Tag (or just its *name*-attribute) that
-            identifies the submit element to use. If ``None``, will choose the
-            first valid submit element in the form, if one exists.
+        :param submit: The :class:`bs4.element.Tag` (or just its
+            *name*-attribute) that identifies the submit element to use. If
+            ``None``, will choose the first valid submit element in the form,
+            if one exists. If ``False``, will not use any submit element;
+            this is useful for simulating AJAX requests, for example.
```
hemberger, line 313: Would you mind extending the documentation of the `False` case to the `btnName` argument of `stateful_browser.submit_selected`?

hemberger (changes requested):
This looks like it works as expected now, thanks for the fixes! I have two very minor suggestions, but once those are addressed I will be happy to approve this PR. Thanks again for the submission!

mechanicalsoup/form.py
alexreg: Sure. I had actually left it in initially, then removed it for economy's sake, but it's probably helpful to have it in there, indeed.

alexreg:
@hemberger I've now expanded the docs for that method (and even reworked it slightly). Hope that's alright.

hemberger (approved):
LGTM, thanks again for your contribution! If there are no other comments from other maintainers, I'll merge this in a day or so.

PR #415, opened 2023-01-23
python cleanup for 3.11

marksmayo:
- converted strings to f-strings
 - sorted imports
 - added trailing commas (as per 3.5, 3.6+)
 - fixed some docstrings
 - fixed spacings
 - made some quotes consistent


tests/test_browser.py
```diff
@@ -3,7 +3,7 @@
 import tempfile
 
 import pytest
-import setpath  # noqa:F401, must come before 'import mechanicalsoup'
+# import setpath  # noqa:F401, must come before 'import mechanicalsoup'
```
hemberger, line 6: The `setpath` import needs to be retained in all the test files. Maybe there's a better way to set the python path for testing these days, but I'm not sure what what would be.

hemberger (changes requested):
Thanks for the PR! For my reference, is this the Black autoformatter?

docs/conf.py
```diff
@@ -16,19 +16,17 @@
...
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 sys.path.insert(0, os.path.abspath('..'))
 
 import mechanicalsoup
 
-
 # -- General configuration ------------------------------------------------
 
 # If your documentation needs a minimal Sphinx version, state it here.
-#needs_sphinx = '1.0'
```
moy, line 31: I disagree with this series of changes. I like the distinction between comments (`# like this`) and commented-out code (`#like = that`). No strong opinion, but I'd rather not change things I'm already happy with.

moy (changes requested):
I'm normally reluctant to this kind of PR, since it introduces a lot of changes (hence potential Git conflicts) with little added value. But MechanicalSoup is not that active these days, so the probability that another change clashes with this one is low, and we try to be as close as possible to perfection (;-) ) and this patch goes in the right direction.

I didn't have time to make a proper review, though. At least the point raised by Dan needs to be fixed.

marksmayo:
> Thanks for the PR! For my reference, is this the Black autoformatter?

I used isort to sort imports, and flake8/pylint to look for issues.

PR #424, opened 2023-11-24
Add support for Python 3.12

hemberger:


moy:
Build breakage seems to be a known problem with flake8 and its dependencies : https://github.com/python/importlib_metadata/issues/406

The simplest for us is probably to disable flake8 for Python >= 3.12, and hope the problem will resolve itself before we actually need it.

hemberger:
I think we're unlikely to see compatibility between pytest and flake8 moving forward (see https://github.com/tholo/pytest-flake8/issues/98). One thing to do would be to run flake8 independently of pytest (I thought I saw that one of the pytest devs recommended this, but I can't find the link anymore). Another alternative is to switch to a different style checker, such as black (https://github.com/psf/black).

moy:
> I think we're unlikely to see compatibility between pytest and flake8 moving forward (see [tholo/pytest-flake8#98](https://github.com/tholo/pytest-flake8/issues/98)).

I didn't see that, but indeed, that's a good reason to move away from pytest-flake8.

But from what I've found on the web, the issue seems to be flake8+python12, not pytest-flake8, so 

> run flake8 independently of pytest

I'm not sure this will solve our issue.

> Another alternative is to switch to a different style checker, such as black (https://github.com/psf/black).

AFAIK, black is a formatter, not a "checker" (IOW, its role is to modify the code to make it conformant with a style checker).

But flake8 is a bit more than a "style" checker. Under the hood it runs pycodestyle (formerly pep8, hence the 8 in the name "flake8") and pyflakes which finds more semantic issues (not as advanced as a real typechecker, but finds stuff like undeclared variables or functions, so typically finds typo in the code), and I've just noticed it also runs "Ned Batchelder's McCabe script" which if I understand correctly computes the cyclomatic complexity.

So, I vote for:

1) run flake8 as standalone tool
2) if it doesn't solve the build issue, disable it for Python 3.12 until the problem is solved in flake8

hemberger:
Your suggestion sounds good to me! If you feel inspired, please feel free to implement this, otherwise I will revisit it when I have time. Thanks!

hemberger:
Superseded by #446.

PR #429, opened 2023-12-20
Pin certifi to 2022.12.7 in production requirements.txt

moy:
This was done by synk in 914958de0cbc4ce3133a54efc73d2cffc43752bc to avoid vulnerability, but for some reasons it was not done in the main requirements.txt.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,5 @@
 requests >= 2.22.0
 beautifulsoup4 >= 4.7
 lxml
+certifi>=2022.12.7  # not directly required (indirect dependency from requests),
+                    # pinned by Snyk to avoid a vulnerability


hemberger (approved):
This seems fine to me, though I wish I understood why Snyk decided it wasn't a vulnerability in our main `requirements.txt` file.

hemberger:
Thanks for taking care of this!

moy:
> This seems fine to me, though I wish I understood why Snyk decided it wasn't a vulnerability in our main `requirements.txt` file.

My guess is that Snyk applied an algorithm like 1) find whatever requirements.txt file in the repo 2) analyze it. It doesn't know semantically which is a test and which is a production file (and indeed it shoud probably have analyzed both).

PR #430, opened 2023-12-20
Actually deal with comments in requirements.txt

moy:
The real solution would be to move to a more modern packaging system like Poetry, but waiting for this to happen we should at least parse properly our own requirement file.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,11 +6,15 @@ from os import path
 from setuptools import setup  # Always prefer setuptools over distutils
 
 
+def strip(line):
+    """Strip comments and whitespace from a line of text."""
+    return line.split('#', 1)[0].strip()
+
+
 def requirements_from_file(filename):
     """Parses a pip requirements file into a list."""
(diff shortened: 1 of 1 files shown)


hemberger:
Thanks! Do we want to retain the removal of `--` suffixes? For example, it is valid to have requirement lines like
```
FooProject >= 1.2 --global-option="--no-user-cfg"
```
(as per https://pip.pypa.io/en/stable/reference/requirements-file-format/) though we obviously don't have any such lines at the moment.

moy:
> Do we want to retain the removal of -- suffixes?

If I read correctly, we never had such *suffix* management, only the removal of lines *starting with* `--`, which was even less useful.

I think we really need to migrate to a more modern tool and abandon our hand-written `requirements.txt` parser, so the less we have in our custom parser the best it is IMHO.

hemberger (approved):
Oh yeah, good point. Thanks!

PR #461, opened 2025-09-21
Refactor form submission

### LINKED ISSUE #230: Action attribute not properly handled when selecting form

hemberger:
The `Form.choose_submit` method is no longer destructive, and instead of removing unselected inputs, it simply stores (and returns) which submit was chosen. This allows the `Form` object to be reused for multiple form submissions, and thus the logic for extracting the input data from the form is fully encapsulated in `Browser.get_request_kwargs`.

Previously, `Browser.get_request_kwargs` assumed that the form data was already processed (e.g. so that there was only one submit/button input), which could result in a malformed request if this was not done by, e.g., calling `Browser.submit` directly. It now has an optional `submit` argument and can correctly process the raw form that may have multiple submit/button inputs.

We add a `StatefulBrowser.submit` method override that only raises an Exception to prevent calling the `Browser.submit` method directly on a StatefulBrowser object. This is an important restriction, because it could be confused with the recommended `StatefulBrowser.submit_selected` method. If needed for some arcane reason, you can pass a StatefulBrowser object into the `Browser.submit` function to recover this behavior.

To ensure that form submission is valid whether we call `Browser.submit` or `StatefulBrowser.submit_selected`, the `Form.choose_submit` call has been moved into `Browser.submit`, which now has a `btnName` argument so that the same argument from `StatefulBrowser.submit_selected` can be passed in.

Please note that `StatefulBrowser.submit_selected` will now override a previously chosen submit instead of raising an Exception if `btnName` is specified after calling `Form.choose_submit` (or an earlier call to `StatefulBrowser.submit_selected` on the same page, which was previously not possible).

Closes #233 (superseded).
Fixes #230.

### DIFF
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -181,16 +181,15 @@ class Browser:
         return request_kwargs
 
     @classmethod
-    def get_request_kwargs(cls, form, url=None, **kwargs):
+    def get_request_kwargs(cls, form, url=None, submit=None, **kwargs):
         """Extract input data from the form."""
         method = str(form.get("method", "get"))
         action = form.get("action")
 
         # If the form has a submit button, use its form action
         # https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#formaction.
-        button_submit_element = form.find("button")
-        if button_submit_element:
-            action = button_submit_element.get("formaction", action)
+        if isinstance(submit, bs4.element.Tag) and (submit.name == "button"):
+            action = submit.get("formaction", action)
 
         url = urllib.parse.urljoin(url, action)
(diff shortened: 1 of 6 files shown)


hemberger:
@moy Just wondering if you have any thoughts about this change. Thanks!

moy:
No time for a detailed review, but from a quick glance it looks good.

PR #466, opened 2026-02-18
Add Form.__getitem__, __iter__, and keys() methods

### LINKED ISSUE #444: Get the fields that would be submitted with a form without actually submitting the form
This would be useful for testing. Right now the only official way to see anything about a `mechanicalsoup.Form` object appears to be `print_summary`, which produces HTML as a string. `__setitem__` is provided, but not `__getitem__`, `__iter__`, or `keys`.

danishashko:
Fixes #444.

`Form.__setitem__` allowed setting values by name, but there was no way to read them back. This adds:

**`__getitem__(name)`** - returns the current field value:
- Text/password/hidden inputs: the `value` attribute.
- Checkboxes: a list of checked values.
- Radio buttons: the selected value, or `None` if none is selected.
- Single `<select>`: the selected option value.
- `<select multiple>`: a list of selected option values.
- `<textarea>`: the text content.
- Raises `KeyError` if the field does not exist.

**`__iter__()`** - yields unique field names in document order.

**`keys()`** - returns a list of unique field names in document order.

Example:
```python
browser.select_form()
browser["size"] = "medium"
assert browser.form["size"] == "medium"
assert "size" in browser.form.keys()
```

Tests for all cases are added in `tests/test_form.py`.

### DIFF
--- a/mechanicalsoup/form.py
+++ b/mechanicalsoup/form.py
@@ -241,6 +241,88 @@ class Form:
         """
         return self.set(name, value)
 
+    def __getitem__(self, name):
+        """Get the current value of the named form field.
+
+        Returns the value that would be submitted for the given field.
+        For checkboxes, returns a list of checked values.
+        For radio buttons, returns the selected value or ``None`` if none
+        is selected.
+        Raises ``KeyError`` if no field with the given name exists.
+
+        Example: ::
+
+            browser.select_form()
+            browser["username"] = "ada"
+            assert browser.form["username"] == "ada"
+        """
+        # Case-insensitive search for type=checkbox
+        checkboxes = self.form.select(f'input[type="checkbox" i][name="{name}"]')
+        if checkboxes:
+            return [cb.get("value", "on") for cb in checkboxes if "checked" in cb.attrs]
+
+        # Case-insensitive search for type=radio
+        radios = self.form.select(f'input[type="radio" i][name="{name}"]')
+        if radios:
+            for r in radios:
+                if "checked" in r.attrs:
+                    return r.get("value", "on")
+            return None
+
+        # Regular input (text, password, hidden, email, etc.)
+        inp = self.form.find("input", {"name": name})
+        if inp:
+            return inp.get("value", "")
+
+        # Select element
+        select = self.form.find("select", {"name": name})
+        if select:
+            if "multiple" in select.attrs:
+                return [
+                    o.get("value", o.get_text())
+                    for o in select.find_all("option")
+                    if "selected" in o.attrs
+                ]
+            selected = select.find("option", attrs={"selected": True})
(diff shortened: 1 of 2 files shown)


hemberger:
@danishashko Thanks for your contribution! I'll try to look more closely at the changes as soon as I can. If there's any specific timeline you're trying to target, just let me know.

mechanicalsoup/form.py
```diff
@@ -241,6 +241,88 @@ def __setitem__(self, name, value):
...
+        Example: ::
+
+            browser.select_form()
+            browser["username"] = "ada"
+            assert browser.form["username"] == "ada"
+        """
+        # Case-insensitive search for type=checkbox
+        checkboxes = self.form.select(f'input[type="checkbox" i][name="{name}"]')
+        if checkboxes:
+            return [cb.get("value", "on") for cb in checkboxes if "checked" in cb.attrs]
```
moy: Shouldn't this be a tuple `(...)` rather than a list `[...]` ? It's not meant to be modified by the caller.

moy (commented):
Not a real review, but on overall this looks good.

hemberger:
@danishashko Would you be willing to fix the few flake8 test failures (line too long) and add whatever coverage test is needed to get back to 100%? If you would prefer that we make these final few changes, that's fine too, just let us know. Thanks again!

danishashko:
Hey @hemberger — done! Just pushed a fix that addresses the flake8 line-length failures and adds a test for the missing coverage (select element defaulting to the first option when none is explicitly selected). Should be back to 100% now. Let me know if anything else needs tweaking!

hemberger:
@danishashko Sorry for the delay in resolving this. Thank you for your latest fixes!

I've been thinking more about this PR, specifically this line in the docstring of `Form.__getitem__`:
```
Returns the value that would be submitted for the given field.
```
I was wondering: if the intent is to be able to ascertain the values of a form element exactly as it would be submitted, it seems like we should be reusing the logic in `Browser.get_request_kwargs`. Otherwise, it would be difficult to guarantee that the queried value is the same value that would be submitted.

I don't have a great idea for how this would be implemented. In `Browser.get_request_kwargs`, we accumulate all submitted values, and we probably want to keep that design for efficiency. However, then that would make `Form.__getitem__` inefficient if it were to share the same implementation (to ensure consistency). Would we cache the results? Then we'd have to be careful about invalidating the cache if the form changes.

Maybe rather than sharing one function to accumulate all submitted values, we could have `Form` methods for querying each type of element (checkbox, radio, input, etc.), and then `Browser.get_request_kwargs` could call each of them as it loops over the elements? In an implementation like this, we would probably want a test that checks for consistency between submission and form querying.

Does anyone else have an opinion on this issue?

PR #473, opened 2026-06-20
Avoid mutable default arguments

duriantaco:
## Summary

Fixed shared default dict in browser methods. Python reuses the same default dict every time a method is called. If that dictionary gets changed, a later call may see the old change. We avoid that by using `None` as the default and then create a new dictionary inside the method.

What we updated:

  - `Browser.__init__` -> avoids sharing the default `soup_config`
  - `StatefulBrowser.follow_link` -> avoids sharing default `bs4_kwargs` and `requests_kwargs`
  - `StatefulBrowser.download_link` -> avoids sharing default `bs4_kwargs` and `requests_kwargs`

## Evidence

  Skylos ([GitHub](https://github.com/RedParrotBerkeley/skylos)) flagged these as `SKY-L001` mutable-default
  findings:

  ```
  mechanicalsoup/browser.py:39
  Mutable default argument detected. This causes state leaks between calls.

  mechanicalsoup/stateful_browser.py:364
  Mutable default argument detected. This causes state leaks between calls.

  mechanicalsoup/stateful_browser.py:365
  Mutable default argument detected. This causes state leaks between calls.

  mechanicalsoup/stateful_browser.py:393
  Mutable default argument detected. This causes state leaks between calls.

  mechanicalsoup/stateful_browser.py:394
  Mutable default argument detected. This causes state leaks between calls.
```

This is manually reviewed, and not an auto bulk rewrite. No AI was used in this cleanup or the PR. 

## Checks

  - pytest
  - skylos . -a --format json
      - Before: 5 SKY-L001 findings
      - After: 0 SKY-L001 findings

### DIFF
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -36,7 +36,7 @@ class Browser:
     :param user_agent: Set the user agent header to this value.
 
     """
-    def __init__(self, session=None, soup_config={'features': 'lxml'},
+    def __init__(self, session=None, soup_config=(('features', 'lxml'),),
                  requests_adapters=None,
                  raise_on_404=False, user_agent=None):
 
@@ -51,7 +51,7 @@ class Browser:
             for adaptee, adapter in requests_adapters.items():
                 self.session.mount(adaptee, adapter)
 
-        self.soup_config = soup_config or dict()
+        self.soup_config = dict(soup_config or ())
 
     @staticmethod
     def __looks_like_html(response):
--- a/mechanicalsoup/stateful_browser.py
+++ b/mechanicalsoup/stateful_browser.py
@@ -361,8 +361,8 @@ class StatefulBrowser(Browser):
                 self.launch_browser()
             raise
 
-    def follow_link(self, link=None, *bs4_args, bs4_kwargs={},
-                    requests_kwargs={},  **kwargs):
+    def follow_link(self, link=None, *bs4_args, bs4_kwargs=(),
+                    requests_kwargs=(),  **kwargs):
         """Follow a link.
 
         If ``link`` is a bs4.element.Tag (i.e. from a previous call to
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: mechanicalsoup/browser.py [hunk 1]
```diff
@@ -51,6 +51,8 @@ def __init__(self, session=None, soup_config={'features': 'lxml'},
             for adaptee, adapter in requests_adapters.items():
                 self.session.mount(adaptee, adapter)
 
+        if soup_config is None:
+            soup_config = {'features': 'lxml'}
         self.soup_config = soup_config or dict()
```
moy: I'd rather have a simpler if/else to avoid an indirection:

```
if soup_config is None:
    self.soup_config = { ... }
else:
    self.soup_config = dict()
```

moy (approved):
A suggestion in comment, but with or without it, the PR looks good to me. Even a test, perfect!

hemberger:
Thanks for the PR! I agree with fixing the mutable default argument issue, however I'm not a fan of hiding default argument values behind `None` placeholders. In python 3.15+, perhaps `frozendict` would be ideal, assuming that we wouldn't need to change the `soup_config` after the Browser is constructed. In the meantime, I'm not sure what the best solution is. Use a 3rd party frozendict library? Use immutable tuples and convert to dict inside the function?

duriantaco:
Wokie, I switched this to immutable tuple defaults and convert them to fresh dicts inside the methods..

mechanicalsoup/browser.py
duriantaco: Changed this in the latest push to use immutable tuple defaults.. U still prefer an explicit if/else here, or is the current dict(...) conversion okay

hemberger (approved):
Thanks for the update! This looks good to me now, but I'm also happy with any further adjustments that @moy might request.

mechanicalsoup/stateful_browser.py
```diff
@@ -392,15 +392,18 @@ def follow_link(self, link=None, *bs4_args, bs4_kwargs={},
 
         :return: Forwarded from :func:`open_relative`.
         """
+        bs4_kwargs = dict(bs4_kwargs or ())
+        requests_kwargs = dict(requests_kwargs or ())
```
moy, line 396: Is this "or ()" really needed? request_kwargs isn't meant to be None...

hemberger: Good point, I don't think it is. I'll do a followup PR.

PR #476, opened 2026-06-22
docs: explain valid features values for soup_config parameter

### LINKED ISSUE #433: Documentation: explain how the BeautifulSoup parser can be specified
In the following part of the documentation, https://github.com/MechanicalSoup/MechanicalSoup/blob/16238f4b726a09563150a2a818cded51ac7d7c1f/mechanicalsoup/stateful_browser.py#L29-L35

...it's mentioned that it's recommended to specify a parser, but looking at BeautifulSoup's documentation, the connection between the `features` and how the parser is specified is not evident because the BS doc never uses the `features` keyword, rather just tosses it in as the second argument:
image
(https://beautiful-soup-4.readthedocs.io/en/latest/#differences-between-parsers)

BS4's documentation also doesn't include an API reference that would list the different parameters, it only provides a [high-level intro](https://beautiful-soup-4.readthedocs.io/en/latest/#beautifulsoup).

I would recommend modifying the MechanicalSoup docs to explain the parser setting more explicitly, like how BS4 does it in code:
```
:param features: Desirable features of the parser to be
         used. This may be the name of a specific parser ("lxml",
         "lxml-xml", "html.parser", or "html5lib") or it may be the
         type of markup to be used ("html", "html5", "xml"). It's
         recommended that you name a specific parser, so that
         Beautiful Soup gives you the same results across platforms
         and virtual environments.
```

highoncomputers:
## Description

The `soup_config` parameter documentation in `browser.py` now explicitly lists all valid parser names and markup types that can be passed as the `features` argument to BeautifulSoup.

## Changes

| File | Change |
|------|--------|
| `mechanicalsoup/browser.py:25-31` | Expanded docstring to list parser names and markup types |

## Before vs After

Before: Generic "specify a parser" with a link to BS4 docs.
After: Lists specific parser values (`lxml`, `lxml-xml`, `html.parser`, `html5lib`) and markup types (`html`, `html5`, `xml`).

## Type of Change
- [x] Documentation

## Related Issue
Fixes #433

### DIFF
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -24,11 +24,17 @@ class Browser:
         constructing a new one.
     :param soup_config: Configuration passed to BeautifulSoup to affect
         the way HTML is parsed. Defaults to ``{'features': 'lxml'}``.
-        If overridden, it is highly recommended to `specify a parser
-        <https://www.crummy.com/software/BeautifulSoup/bs4/doc/#specifying-the-parser-to-use>`__.
-        Otherwise, BeautifulSoup will issue a warning and pick one for
-        you, but the parser it chooses may be different on different
-        machines.
+        If overridden, it is highly recommended to specify a parser
+        explicitly. The ``features`` argument to BeautifulSoup may be
+        the name of a specific parser (``"lxml"``, ``"lxml-xml"``,
+        ``"html.parser"``, or ``"html5lib"``), or the type of markup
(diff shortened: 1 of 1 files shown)


moy:
This (and other PR by the same author) looks sensible. Why is it closed?

hemberger:
> This (and other PR by the same author) looks sensible. Why is it closed?

They deleted their fork. I'm guessing it was some kind of AI PR creator.

PR #478, opened 2026-06-26
Add return typehint for ResponseWithSoup objects

hemberger:
Begin to add typehint infrastructure by solving the biggest limiting factor: mutating the type of the Response object returned by requests. While it may be better to use an intersection between a Response object and a Protocol that has a `soup` attribute, intersection is not yet well-supported in the typing ecosystem. Instead, use a class derived from the requests Response object that contains a `soup` attribute and explicitly mutate the type in the `add_soup` method.

Add a mypy GitHub Actions test to validate any types that we specify.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -13,7 +13,6 @@ on:
 
 jobs:
   build:
-
     runs-on: ubuntu-24.04
     strategy:
       fail-fast: false
@@ -53,3 +52,14 @@ jobs:
         fail_ci_if_error: true
         token: ${{ secrets.CODECOV_TOKEN }}
         verbose: true
+
+  mypy:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v6
+      - name: Install dependencies
+        run: |
+            python -m pip install --upgrade pip
+            python -m pip install mypy -r requirements.txt
+      - name: Run mypy
+        run: mypy mechanicalsoup
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -60,6 +60,6 @@ jobs:
       - name: Install dependencies
         run: |
             python -m pip install --upgrade pip
-            python -m pip install mypy -r requirements.txt
+            python -m pip install -r requirements.txt -r tests/requirements.txt
       - name: Run mypy
         run: mypy mechanicalsoup
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -1,6 +1,7 @@
 import io
 import os
 import tempfile
+from typing import cast
 import urllib
 import weakref
 import webbrowser
@@ -12,6 +13,10 @@ import requests
 from .__version__ import __title__, __version__
 from .form import Form
 from .utils import LinkNotFoundError, is_disabled, is_multipart_file_upload
+
+
+class ResponseWithSoup(requests.Response):
+    soup: bs4.BeautifulSoup | None
 
 
 class Browser:
@@ -36,10 +41,14 @@ class Browser:
     :param user_agent: Set the user agent header to this value.
 
     """
-    def __init__(self, session=None, soup_config=(('features', 'lxml'),),
-                 requests_adapters=None,
-                 raise_on_404=False, user_agent=None):
-
+    def __init__(
+        self,
+        session: requests.Session | None = None,
+        soup_config=(('features', 'lxml'),),
+        requests_adapters=None,
+        raise_on_404=False,
+        user_agent=None,
+    ):
         self.raise_on_404 = raise_on_404
         self.session = session or requests.Session()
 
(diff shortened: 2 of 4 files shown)


hemberger:
@moy Once we're happy with this, I'll merge and start fleshing out more of the typehints.

hemberger:
@moy Just wanted to check one last time if you have any opinion on this approach before I merge. After this, I'd like to get some basic typehinting in place and then do a 1.5 release. Thanks! :)

moy:
No strong opinion on which approach is the best in practice. Did you try pyright also? In my experience it gives better diagnosis than MyPy, but my experience is in a very limited and specific context (student compiler project including both generated code and external dependencies with broken typing). But playing with two different tools may help you identify weakness in the approach.

hemberger:
Yep, I think validating against both mypy and pyright will be useful, but there are some errors with pyright with the existing codebase that will require some additional work. Some are easy to fix, but others look like they might be false positives. Here's what I see right now:
```
$ pyright mechanicalsoup
mechanicalsoup/browser.py
  mechanicalsoup/browser.py:115:22 - error: "cookies" is not a known attribute of "None" (reportOptionalMemberAccess)
  mechanicalsoup/browser.py:119:29 - error: "cookies" is not a known attribute of "None" (reportOptionalMemberAccess)
  mechanicalsoup/browser.py:129:22 - error: "headers" is not a known attribute of "None" (reportOptionalMemberAccess)
  mechanicalsoup/browser.py:144:33 - error: "request" is not a known attribute of "None" (reportOptionalMemberAccess)
  mechanicalsoup/browser.py:155:33 - error: "get" is not a known attribute of "None" (reportOptionalMemberAccess)
  mechanicalsoup/browser.py:168:33 - error: "post" is not a known attribute of "None" (reportOptionalMemberAccess)
  mechanicalsoup/browser.py:179:33 - error: "put" is not a known attribute of "None" (reportOptionalMemberAccess)
  mechanicalsoup/browser.py:202:22 - error: "parse" is not a known attribute of module "urllib" (reportAttributeAccessIssue)
  mechanicalsoup/browser.py:326:29 - error: "request" is not a known attribute of "None" (reportOptionalMemberAccess)
mechanicalsoup/stateful_browser.py
  mechanicalsoup/stateful_browser.py:135:23 - error: "parse" is not a known attribute of module "urllib" (reportAttributeAccessIssue)
  mechanicalsoup/stateful_browser.py:187:29 - error: "send" is not a known attribute of "None" (reportOptionalMemberAccess)
  mechanicalsoup/stateful_browser.py:226:31 - error: "find_all" is not a known attribute of "None" (reportOptionalMemberAccess)
  mechanicalsoup/stateful_browser.py:236:37 - error: "select" is not a known attribute of "None" (reportOptionalMemberAccess)
  mechanicalsoup/stateful_browser.py:247:23 - error: Argument of type "Literal['id']" cannot be assigned to parameter "key" of type "SupportsIndex | slice[SupportsIndex | None, SupportsIndex | None, SupportsIndex | None]" in function "__getitem__"
    Type "Literal['id']" is not assignable to type "SupportsIndex | slice[SupportsIndex | None, SupportsIndex | None, SupportsIndex | None]"
      "Literal['id']" is incompatible with protocol "SupportsIndex"
        "__index__" is not present
      "Literal['id']" is not assignable to "slice[SupportsIndex | None, SupportsIndex | None, SupportsIndex | None]" (reportArgumentType)
  mechanicalsoup/stateful_browser.py:325:31 - error: "find_all" is not a known attribute of "None" (reportOptionalMemberAccess)
  mechanicalsoup/stateful_browser.py:406:35 - error: Object of type "None" is not subscriptable (reportOptionalSubscript)
  mechanicalsoup/stateful_browser.py:429:33 - error: Object of type "None" is not subscriptable (reportOptionalSubscript)
  mechanicalsoup/stateful_browser.py:433:33 - error: "get" is not a known attribute of "None" (reportOptionalMemberAccess)
18 errors, 0 warnings, 0 informations
```
The issue with thinking that `self.session` is None, for example, doesn't seem to have an obvious fix (and I've tried restating the definition in the initializer). I'm inclined to move forward with mypy for now, while continuing to consult with pyright locally until all of its errors can be addressed.

moy:
Sometimes, adding some "assert foo is not None", or even "assert isinstance(foo, bar)" can help. Somehow it turns back the static typing into a runtime check, but it does kill the static typing error. The typing is flow sensitive so the type checker propagates the information below the assertion.

hemberger:
I've made more progress on fixing some of the pyright errors, but I would still like to make the changes incrementally as much as possible. The changes needed to enable pyright raise an entirely different set of questions, so I think that's suitable for a followup PR. Are you okay with me taking that approach and merging this PR, or would you rather have a more comprehensive single PR that addresses all of the complexities in converting to a typed codebase?

moy:
> The changes needed to enable pyright raise an entirely different set of questions, so I think that's suitable for a followup PR

No problem for me with a separate PR. I'd just advise to play a bit with Pyright before merging the mypy one, just to check that you're not missing an important issue. If you're confident that the Pyright stuff can be fixed without re-doing the mypy one, then perfect.

PR #479, opened 2026-07-07
Do not submit controls inside a disabled <fieldset>

gaoflow:
## Problem

`Browser.get_request_kwargs` skips a form control from submission only when the
control carries its own `disabled` attribute:

```python
if tag.has_attr('disabled'):
    continue
```

Per the [HTML specification](https://html.spec.whatwg.org/multipage/form-elements.html#concept-fieldset-disabled),
a control is *also* disabled when it descends from a `<fieldset>` whose
`disabled` attribute is set — except when it is inside that fieldset's first
`<legend>`, which stays enabled — and disabled controls are barred from
submission. MechanicalSoup, which aims for browser fidelity, submitted them
anyway.

Given a form with a `<fieldset disabled>`, a real browser submits only the
controls outside the fieldset (and those in its first `<legend>`), whereas
MechanicalSoup submitted every control:

```python
# emitted: {'enabled': '1', 'in_legend': 'legend', 'text': '2',
#           'check': 'on', 'sel': 'x', 'area': 'hi', 'nested': '3'}
# browser: {'enabled': '1', 'in_legend': 'legend'}
```

## Fix

Add an `is_disabled(tag)` helper that returns `True` for a control with its own
`disabled` attribute *or* one descending from a disabled ancestor `<fieldset>`,
honoring the first-`<legend>` exemption and nested-fieldset propagation, and use
it in place of the bare `has_attr('disabled')` check. This extends the existing
disabled-control handling (from #248) rather than changing behavior introduced
by the #461 refactor.

## Tests

`tests/test_browser.py::test__request_disabled_fieldset` builds a form with a
disabled fieldset containing text/checkbox/select/textarea/button controls, a
first `<legend>`, and a nested inner fieldset, then asserts only the controls
outside the fieldset and inside its first `<legend>` are submitted. It fails
before the change and passes after. Full suite passes (124 passed;
`test_launch_browser` errors on pristine HEAD too — an unrelated headless
`webbrowser` launch). `flake8` clean.

### DIFF
--- a/docs/ChangeLog.rst
+++ b/docs/ChangeLog.rst
@@ -5,6 +5,10 @@ Release Notes
 Version 1.5 (in development)
 ============================
 
+* Form controls inside a disabled ``<fieldset>`` are no longer submitted,
+  matching browser behavior and the HTML specification. Controls inside the
+  fieldset's first ``<legend>`` remain enabled.
+
 Version 1.4
 ===========
 
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -11,7 +11,7 @@ import requests
 
 from .__version__ import __title__, __version__
 from .form import Form
-from .utils import LinkNotFoundError, is_multipart_file_upload
+from .utils import LinkNotFoundError, is_disabled, is_multipart_file_upload
 
 
 class Browser:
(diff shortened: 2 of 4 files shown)


hemberger (approved):
This looks to correctly implement the part of the spec about disabled fieldsets. Thanks for your contribution!

I'll merge this in a few days to give other maintainers an opportunity to review.

PR #483, opened 2026-07-25
Raise an explanatory error from links() when no page is loaded

### LINKED ISSUE #410: links() should return an empty list if self.page is None
I was writing a fuzzer for a cybersecurity assignment, and it crashed when it tried to find the links on a PDF file. I think it would make more sense to return that there are no links, if the page fails to parse. This seems relatively straightforward to implement.

Sanjays2402:
Closes #410

`links()` called `self.page.find_all()` unconditionally, so calling it before any page is opened — or after a response that wasn't parsed as a page (the reporter hit a PDF) — failed with `'NoneType' object has no attribute 'find_all'`, which gives no hint about the real cause.

Per @hemberger's suggestion on the issue, this keeps the exception (rather than returning `[]`, which would be indistinguishable from a page with no links) but makes the message explain the condition, mirroring how the `form` property already reports an unselected form.

### DIFF
--- a/docs/ChangeLog.rst
+++ b/docs/ChangeLog.rst
@@ -8,6 +8,8 @@ Version 1.5 (in development)
 * Form controls inside a disabled ``<fieldset>`` are no longer submitted,
   matching browser behavior and the HTML specification. Controls inside the
   fieldset's first ``<legend>`` remain enabled.
+* ``StatefulBrowser.links`` now raises an explanatory ``AttributeError`` when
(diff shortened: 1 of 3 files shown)


hemberger (approved):
Thanks for your contribution, @Sanjays2402! I'm happy with this solution because it keeps the behavior the same while providing a more user-friendly error message. This will also make it easier to decide in the future if we want to change the behavior.

moy (approved):
Ideally there couls be one more test with a non-HTML page, but that may be overkill.

PR #484, opened 2026-07-25
Detect HTML from raw leading bytes in __looks_like_html

### LINKED ISSUE #432: Suggestions for improvement on this time-consuming function: __looks_like_html()
In practical, it has been observed that this function can be very time-consuming if encountering image files or other larger binary files. It is recommended to modify it to this or similars:
```
def __looks_like_html(response):
        """Guesses entity type when Content-Type header is missing.
        Since Content-Type is not strictly required, some servers leave it out.
        """
        #text = response.text.lstrip().lower()
        #return text.startswith('<html') or text.startswith('<!doctype')
        return re.search(br'<html|<!doctype', response.content[:200]) is not None
```

https://github.com/MechanicalSoup/MechanicalSoup/blob/91b1207a7a2e02233b98b7c2a2a49c3fdaa492de/mechanicalsoup/browser.py#L62

Sanjays2402:
Fixes #432

`__looks_like_html` decoded the entire response body via `response.text` (which also runs charset sniffing) just to inspect its first few characters, which is slow for large binary responses such as images or PDFs. It now matches a compiled bytes regex against the first 200 bytes of `response.content`, with `re.IGNORECASE` as suggested in the issue — so uppercase markup like `<HTML>` is detected too, which the previous `lower()` only handled after a full decode. NUL bytes are stripped for a second attempt so UTF-16 markup is still recognised.

Tests cover lowercase/uppercase/doctype/UTF-16 markup plus PNG and JSON bodies, and one test asserts `response.text` is never accessed during sniffing (fails on the old implementation, passes on the new one).

### DIFF
--- a/mechanicalsoup/browser.py
+++ b/mechanicalsoup/browser.py
@@ -1,5 +1,6 @@
 import io
 import os
+import re
 import tempfile
 import urllib
 import weakref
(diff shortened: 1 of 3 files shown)


hemberger (approved):
Thanks for another great contribution! This looks reasonable to me: it implements the previous attempt with a few additional bells and whistles (optional byte-order marks, handling UTF-16/UTF-32 cases). It also includes a micro-optimization for the most common pathway where the response is HTML. We could just use the second call to match and omit the first, if we prioritize simplicity over speed. I could go either way on this. I'd be curious to hear @moy's opinion.

mechanicalsoup/browser.py
```diff
@@ -53,13 +54,26 @@ def __init__(self, session=None, soup_config=(('features', 'lxml'),),
 
         self.soup_config = dict(soup_config or ())
 
+    _HTML_MARKER_RE = re.compile(
+        rb'\s*(?:\xef\xbb\xbf|\xff\xfe|\xfe\xff)?\s*(?:<html|<!doctype)',
```
moy, line 60: Please add a comment explaining the hexdecimal stuff (utf8 bom if I understand correctly).

moy (approved):
Didn't check the new tests.

I think the two-calls current version is fine: the reader first reads the normal case, and then only has to consider the unusual utf16/32 cases.

PR #486, opened 2026-07-29
Fix some pyright errors

hemberger:
* Pyright requires explicitly importing submodules, so we import `urllib.parse` instead of just `urllib`.

* Some StatefulBrowser methods would implicitly fail if they were called before we set the internal `__state` due to `self.page` and `self.url` being used in a non-None context when they are None (in `select_form` and `absolute_url` respectively). Make these failures explicit so that Pyright understands the failure is expected.

* Pyright was complaining that the literal string "id" could not be used with `bs4.Tag.__getitem__` for its key argument. This is likely a false positive, but switching to `bs4.Tag.get` fixes the error and has the same return value since we have already checked that the "id" attribute exists.

* In `StatefulBrowser.select_form`, Pyright did not remember that the type of `self.page` was narrowed to be not None inside the subfunction `find_associated_elements`. To fix this, simply pass the narrowed `self.page` as an argument to the subfunction.

* The `StatefulBrowser._find_link_internal` method cannot return None, but Pyright seemed to infer that the first conditional could return None. Add an innocuous explicit check against None to make this clear to Pyright.

### DIFF
--- a/mechanicalsoup/stateful_browser.py
+++ b/mechanicalsoup/stateful_browser.py
@@ -1,6 +1,6 @@
 import re
 import sys
-import urllib
+import urllib.parse
 
 import bs4
 import requests
@@ -132,6 +132,11 @@ class StatefulBrowser(Browser):
         ``url``, as in the `.urljoin() method of urllib.parse
         <https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin>`__.
         """
+        if self.url is None:
+            raise AttributeError(
+                "No page has been opened yet on this browser, or the last "
+                "response was not parsed as a page (e.g. it was not HTML)."
+            )
         return urllib.parse.urljoin(self.url, url)
 
     def open(self, url, *args, **kwargs) -> ResponseWithSoup:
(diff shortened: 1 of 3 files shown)


mechanicalsoup/stateful_browser.py
```diff
@@ -362,7 +371,7 @@ def _find_link_internal(self, link, args, kwargs):
         * If searching for the link fails and debug is active, launch
           a browser.
         """
-        if hasattr(link, 'attrs') and 'href' in link.attrs:
+        if link and hasattr(link, 'attrs') and 'href' in link.attrs:
```
moy, line 374: if `link` cannot be `None`, I'd rather have `assert link is not None` to make it explicit to the reader that we know it's not `None`.

But from a very quick look at the code, it seems `_find_link_internal` is called with a `link` argument which defaults to `None` in `follow_link` and `download_link`, so `link` could legitimately be `None`. If this is the case and if I'm reading the PR's description correctly, then the patch is OK (and fixes a real bug in case `link` is `None`), but the PR's description should be fixed.

hemberger: What I meant to say is that the return value for the function `_find_link_internal` cannot be `None`, and so inside the conditional:
```python
if hasattr(link, 'attrs') and 'href' in link.attrs:
    return link
```
`link` as returned here cannot be `None`. But it is certainly true that `link=None` can be passed as an input argument to `_find_link_internal`.

For reasons that I don't understand, pyright did not register that if the conditional is true, `link` cannot be `None`, so I made it explicit by changing it to:
```python
if link and hasattr(link, 'attrs') and 'href' in link.attrs:
```
This was the only way I could think of to get pyright to narrow `link` inside the conditional. I'm open to alternatives if you see a better approach!

PR #489, opened 2026-09-22
Explain how soup_config selects the HTML parser

### LINKED ISSUE #433: Documentation: explain how the BeautifulSoup parser can be specified

saitakarcesme:
Fixes #433.

Explain in both `Browser` and `StatefulBrowser` docstrings that `soup_config` selects the BeautifulSoup parser through its `features` key. Include a complete `html.parser` configuration, name the `lxml` and `html5lib` alternatives, and clarify which parser needs an installed package.

Validation (macOS, CPython 3.13.5): all three parser configurations were executed with both browser classes against local HTML, checking the selected builder and parsed text. Flake8 on both edited files and `git diff --check` pass. Sphinx 9.1.0 generates HTML, but `-W` exits nonzero on the existing anonymous-link error in `docs/ChangeLog.rst:401`; the unchanged base produces the same single diagnostic. No runtime code changed, and the full network-dependent test suite was not run.

Prepared and validated with AI assistance (OpenAI Codex).


moy (approved):
Looks good to me. Perhaps a link to BeautifulSoup's documentation about parsers can be added.

saitakarcesme:
Thanks for reviewing. Added direct links to BeautifulSoup's parser comparison and differences-between-parsers sections in both Browser and StatefulBrowser docstrings in 69ab098. Verified all four links in the generated HTML; flake8 and git diff --check pass. The full strict Sphinx build still reports only the pre-existing anonymous-link error in docs/ChangeLog.rst:401 noted in the PR.

moy (approved):

